PR #11, opened 2019-12-20
switch repository to bionic

colearendt:
This isn't going to fare so well otherwise 😄


danielfrg:
rofl :D

colearendt:
Also, I didn't realize we're using http there... should we switch that to https? 😆 Probably something worth batching w/ another changes. I don't know how I feel about one character PRs haha

danielfrg:
+1 for 1 character PRs

PR #19, opened 2019-12-27
Add RSPM R

### LINKED ISSUE #17: Configure Package Manager to use R
The Package Manager container has R installed but is not configured to use it. I propose we should have the default configuration support R / building packages from git / etc.

### LINKED ISSUE #18: Add Makefile helper for R versions
Similar to what we did for product versions, it will probably be worth adding a `Makefile` helper for when we have to update the R versions (i.e. related to #17). Locations:

- in each Dockerfile
- in `rstudio-pm.gcfg`

colearendt:
This makes RSPM more "full featured." It requires keeping `R_VERSION` in sync, so I made the version explicit and added some Makefile helpers to keep things in sync (following the pattern of the product version numbers).

Close #17 
close #18


danielfrg:
Good stuff, lgtm.

PR #23, opened 2019-12-28
Fix testing

colearendt:
The guidance specified in Docker Hub's docs (https://docs.docker.com/docker-hub/builds/advanced/) suggests to use `build: .`, which re-triggers a build of the `sut` (system under test?)... which we _just built_. Moreover, what you are testing is _not_ the image that you built and are tagging, which is definitely a no-no.

In the `daily` case, we were failing the build even though the image we built is right... the re-trigger was building a bad image. Thankfully that happened instead of the opposite (passing test w/ bad image to be tagged)!

Tsk tsk, DockerHub. I'm hopeful `image: $IMAGE_NAME` will fix the issue and actually test the image we built 😄Added some more auditing just in case we ever have to understand what's going on again.


colearendt:
💰 💰 💰 

Now:
```
Successfully built b1a34515e564
Successfully tagged rstudio/rstudio-server-pro:this
Starting Test in docker-compose.test.yml...
sut uses an image, skipping
```

Versus previously:
```
Successfully built 2b462565fef2
Successfully tagged rstudio/rstudio-server-pro:this
Starting Test in docker-compose.test.yml...
Building sut
Step 1/35 : ARG R_VERSION=3.6
```

PR #29, opened 2020-02-04
Add git and repository

colearendt:
Add git to the connect image
Add RPackageRepository configuration by default (is there a reason we left that off? or just forgot? I know we do it for RSP)

### DIFF
--- a/connect/rstudio-connect.gcfg
+++ b/connect/rstudio-connect.gcfg
@@ -32,3 +32,9 @@ Provider = password
 [Python]
 Enabled = true
 Executable = /opt/python/3.6.5/bin/python
+
+[RPackageRepository "CRAN"]
+URL = https://demo.rstudiopm.com/all/__linux__/bionic/latest
+
+[RPackageRepository "RSPM"]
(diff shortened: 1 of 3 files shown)


slopp:
I'm not sure I love codifying demo here... Could we at least add a nod in the documentation to how that should be changed to your RSPM server? 

Or better yet, add the docs and a comment in the gcfg file

connect/rstudio-connect.gcfg
```diff
@@ -32,3 +32,9 @@ Provider = password
 [Python]
 Enabled = true
 Executable = /opt/python/3.6.5/bin/python
+
+[RPackageRepository "CRAN"]
+URL = https://demo.rstudiopm.com/all/__linux__/bionic/latest
```
danielfrg: Following discussion on slack and here lets comments this lines out.

danielfrg:
Can you also commend this line https://github.com/rstudio/rstudio-docker-products/blob/master/server-pro/conf/repos.conf as part of this PR?

colearendt:
Apologies for the delay here! Good to go?

danielfrg:
LGTM

Issue #27, opened 2020-01-28
Add Instructions for Using a License Server


nicholsn:
Is it currently possible to connect to a license server rather than passing a `License Key`? If so, it would be great if the instructions could be added to the readme.

Also, not sure if this would make sense, but it might be useful to have a docker image added here that could be used to run the license server.

colearendt:
Thanks for commenting here! That is a fantastic point - we definitely should document that! 

Regarding running a docker image for running the license server - we generally recommend against that practice since any "reboot" of the container could cause trouble for all of your licensing. We will think through this a bit more though 😄Do you run the license server in a container today?

nicholsn:
Thanks and I see, that makes sense why you wouldn't run the license server in docker. Our license server is running as a standalone server. Appreciate the offer to add docs!

PR #31, opened 2020-03-19
Floating licensing and launcher off option

### LINKED ISSUE #9: Create ENV var to disable launcher
https://github.com/rstudio/rstudio-docker-products/blob/a7ddff5991edcd247d0e85f3dc6c38e3ba265d78/server-pro/startup.sh#L37-L38

I think to make this container ideal, we should make launcher "optional." A few reasons for this:

- If you want to use legacy behavior without the launcher (i.e. project sharing, load balancing local sessions, etc.)
- If you want to run the launcher in a separate container from RSP. This pattern isn't common yet, but it is something we may want to think about as an option longer term 😄

colearendt:
- Add a `LICENSE_SERVER` env var for each product (related to #27)
- Add a `USE_LAUNCHER` env var for RSP that defaults to `true` and requires `true` to run the launcher (close #9)

- Should probably add some docs. Does this seem like the right approach? Should we do something like `RSPM_LICENSE_SERVER` for each? That seemed like a bit overkill to me.


danielfrg:
How are you testing the license server? and external one on EC2?
This lgtm but it would be ideal if could add a docker container for the license server :D, totally fine if we do it in another PR. I have never done that so i will be happy to take it.

colearendt:
I actually have a license server docker container handy 😄 So super easy to add.

colearendt:
Thoughts on this approach? A `helper/float` directory? I figured the other thing that may live there is a `helper/proxy` or `helper/landing` or some such. Maybe others 😄

Also, thoughts on the `USE_LAUNCHER` env var? Should it just be `LAUNCHER`?

colearendt:
@danielfrg if you don't mind taking another pass at this when you have a chance, that'd be great! 😄

Makefile
```diff
@@ -42,45 +53,59 @@ update-versions:  ## Update the version files for all products
...
 
+
 run-rsp: run-server-pro
 run-server-pro:  ## Run RSP container
 	docker run -it --privileged \
 		-p 8787:8787 \
 		-v /run \
-		-e RSP_LICENSE=$(RSP_LICENSE) \
+		-e LICENSE=$(RSP_LICENSE) \
+		-e LICENSE_SERVER=$(RSP_LICENSE_SERVER) \
```
colearendt, line 68: @danielfrg Why `LICENSE_SERVER` here and not on the others?

colearendt:
@danielfrg Merge conflict w/ Makefile.

Also, do we like the pattern of using `LICENSE` for the containers as well? I'm not sure I'm a huge fan of that pattern... The pattern feels reasonable for the floating license server b/c they are for testing / etc... but running a docker container w/ the product is much easier when I do:
```
docker run -e RSP_LICENSE -p 8787 rstudio/rstudio-server-pro:latest
```

Mapping `LICENSE=RSP_LICENSE` feels a little funky. 😬 Maybe this is just me though

danielfrg:
Oh really? XD. I changed it when I saw you used `LICENSE_SERVER=RSP_LICENSE_SERVER` and i liked it more since the environment variable is the same independent of the container.

I prefer the single `LICENSE` and `LICENSE_SERVER` as variables for the container a bit more but I am ok if we make them consistent, so `RSP_LICENSE` and `RSP_LICENSE_SERVER`, this is also something other container do, to prefix something. If you like that one more I can change.

Makefile
danielfrg: My bad, removed.

colearendt:
Ahh I don't know!! Decisions are hard 😭 The consistency _is_ nice...

danielfrg:
Yeah the consistency makes it nice since we provide multiple images, in cases like this i usually do what postgres or other docker images do and they do `POSTGRES_XXXXX`, so maybe we go with that?

danielfrg:
Things are updated now but git is not showing it here, you can take a final look at the PREFIX and merge if it looks good!

Makefile
```diff
@@ -42,57 +53,74 @@ update-versions:  ## Update the version files for all products
...
 	cd ./server-pro && IMAGE_NAME=rstudio/rstudio-server-pro:$(RSP_VERSION) docker-compose -f docker-compose.test.yml run sut
 
+
 run-rsp: run-server-pro
 run-server-pro:  ## Run RSP container
 	docker run -it --privileged \
 		-p 8787:8787 \
 		-v /run \
-		-e RSP_LICENSE=$(RSP_LICENSE) \
+		-e LICENSE=$(RSP_LICENSE) \
```
colearendt, line 72:
```suggestion
		-e RSP_LICENSE=$(RSP_LICENSE) \
```

Makefile
```diff
@@ -42,57 +53,74 @@ update-versions:  ## Update the version files for all products
...
 
+
 run-rsc: run-connect
 run-connect:  ## Run RSC container
 	docker run -it --privileged \
 		-p 3939:3939 \
 		-v $(CURDIR)/data/rsc:/var/lib/rstudio-connect \
 		-v $(CURDIR)/connect/rstudio-connect.gcfg:/etc/rstudio-connect/rstudio-connect.gcfg \
-		-e RSC_LICENSE=$(RSC_LICENSE) \
+		-e LICENSE=$(RSC_LICENSE) \
```
colearendt, line 92:
```suggestion
		-e RSC_LICENSE=$(RSC_LICENSE) \
```

Makefile
```diff
@@ -42,57 +53,74 @@ update-versions:  ## Update the version files for all products
...
 
+
 run-rspm: run-package-manager
 run-package-manager:  ## Run RSPM container
 	docker run -it --privileged \
 		-p 4242:4242 \
 		-v $(CURDIR)/data/rspm:/data \
 		-v $(CURDIR)/package-manager/rstudio-pm.gcfg:/etc/rstudio-pm/rstudio-pm.gcfg \
-		-e RSPM_LICENSE=$(RSPM_LICENSE)  \
+		-e LICENSE=$(RSPM_LICENSE)  \
```
colearendt, line 112:
```suggestion
		-e RSPM_LICENSE=$(RSPM_LICENSE)  \
```

README.md
```diff
@@ -189,13 +194,15 @@ with a persistent volume in the host machine or your docker orchestration system
...
+2. Setting the `RSPM_LICENSE_SERVER` environment variable to a valid license server / port
+3. Mounting a `/etc/rstudio-pm/license.lic` single file that contains a valid license for RStudio Package Manager
 
 #### Environment variables
 
 | Variable | Description | Default |
 |-----|---|---|
-| `RSPM_LICENSE` | License key for RStudio Package Manager, format should be: `XXXX-XXXX-XXXX-XXXX-XXXX-XXXX-XXXX` | None |
+| `LICENSE` | License key for RStudio Package Manager, format should be: `XXXX-XXXX-XXXX-XXXX-XXXX-XXXX-XXXX` | None |
+| `LICENSE_SERVER` | Floating license server, format should be: `my.url.com:port` | None |
```
colearendt, line 205:
```suggestion
| `RSPM_LICENSE_SERVER` | Floating license server, format should be: `my.url.com:port` | None |
```

README.md
```diff
@@ -189,13 +194,15 @@ with a persistent volume in the host machine or your docker orchestration system
...
-2. Mounting a `/etc/rstudio-pm/license.lic` single file that contains a valid license for RStudio Package Manager
+2. Setting the `RSPM_LICENSE_SERVER` environment variable to a valid license server / port
+3. Mounting a `/etc/rstudio-pm/license.lic` single file that contains a valid license for RStudio Package Manager
 
 #### Environment variables
 
 | Variable | Description | Default |
 |-----|---|---|
-| `RSPM_LICENSE` | License key for RStudio Package Manager, format should be: `XXXX-XXXX-XXXX-XXXX-XXXX-XXXX-XXXX` | None |
+| `LICENSE` | License key for RStudio Package Manager, format should be: `XXXX-XXXX-XXXX-XXXX-XXXX-XXXX-XXXX` | None |
```
colearendt, line 204:
```suggestion
| `RSPM_LICENSE` | License key for RStudio Package Manager, format should be: `XXXX-XXXX-XXXX-XXXX-XXXX-XXXX-XXXX` | None |
```

connect/rstudio-connect.gcfg
```diff
@@ -1,5 +1,8 @@
 ; RStudio Connect configuration file
 
+[Licensing]
+LicenseType = remote
```
colearendt, line 4: We don't want this do we?

connect/startup.sh
```diff
@@ -11,14 +11,17 @@ deactivate() {
 trap deactivate EXIT
 
 # Activate License
-if ! [ -z "$RSC_LICENSE" ]; then
-    /opt/rstudio-connect/bin/license-manager activate $RSC_LICENSE
+if ! [ -z "$RSP_LICENSE" ]; then
```
colearendt, line 14: Eek! RSP variables in the RSC startup file

server-pro/conf/rserver.conf
```diff
@@ -1,3 +1,4 @@
+server-license-type=remote
```
colearendt, line 1: Should probably remove this

danielfrg:
There is a super weird mix between things that are fixed but dont show up on Github, just when you pull and stuff that is actually wrong. I will make one more commit and see if that fixes this state.

danielfrg:
Ok, i think this is in better shape now!

colearendt:
Nice! LGTM!

KVInventoR:
Hi @colearendt , @danielfrg 

Can you please clarify: this PR included a new file: rserver-float.conf 
Does RStudio server pro support a new file rserver-float.conf for configure remote license server?

Or probably: startup.sh should include condition where is RSP_LICENSE_SERVER present, rserver-float.conf should be added to rserver.conf  to the end of file?

colearendt:
@KVInventoR Thanks for reaching out! We should probably make some better documentation on this.

The `rserver-float.conf` is an _alternative_ to `rserver.conf`. (Floating licensing today requires setting `server-license-type=remote` in the config file, as well as activating the license manager in a different way).

https://github.com/rstudio/rstudio-docker-products/blob/3931bea3c5ec767caa28b8a5e2b0bda28a0ad3c5/server-pro/startup.sh#L34

So basically, in order to start RSP with floating licensing:
- set the `RSP_LICENSE_SERVER` env var
- add `server-license-type=remote` to your `rserver.conf` file

Ultimately, I suspect we will not want to monkey with the `rserver.conf` at runtime, lest we modify a mounted file / etc.

I have created an issue to track improvements if you would like to share any follow up discussion / feedback! https://github.com/rstudio/rstudio-docker-products/issues/94

PR #36, opened 2020-03-24
Actually fix touch

colearendt:
Get this... the last PR fixed it _locally_, but not when built on docker hub... WAT.

Something strange is afoot. In any case, I added tests so now builds should fail unless it's actually fixed

### DIFF
--- a/server-pro/startup.sh
+++ b/server-pro/startup.sh
@@ -38,8 +38,7 @@ fi
 wait-for-it.sh localhost:5559 -t $LAUNCHER_TIMEOUT
 
 # touch log files to initialize them
-touch /var/lib/rstudio-server/monitor/log/rstudio-server.log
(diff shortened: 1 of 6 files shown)


server-pro/startup.sh
```diff
@@ -38,8 +38,7 @@ fi
 wait-for-it.sh localhost:5559 -t $LAUNCHER_TIMEOUT
 
 # touch log files to initialize them
-touch /var/lib/rstudio-server/monitor/log/rstudio-server.log
-chown rstudio-server:rstudio-server /var/lib/rstudio-server/monitor/log/rstudio-server.log
+su rstudio-server -c 'touch /var/lib/rstudio-server/monitor/log/rstudio-server.log'
```
danielfrg: lol ok... thanks docker

PR #39, opened 2020-04-06
Fix connect

colearendt:
Worth noting that our 1.8.2 connect image doesn't work :sob: The python runtime breaks against newer Connect versions b/c old setuptools and newer virtualenv. I added a test to catch this in the future by actually invoking Connect.

A few controversial changes:
- "Starting Connect" is the only way I am aware of doing the "runtime checks"
- As a result, and since Connect does not exit, I used a `timeout` and gave Connect 5 seconds to start up and throw errors before calling things good. This is probably a bad idea, but I couldn't come up with anything else
- In the interest of "test driven development," this test would have caught this issue before we bumped the version 😄 
- I added `make test-*-i` targets to start a bash shell in the test container, thereby making reproducing tests easier 😄 
    - After doing this, I realized that `RELEASE.md` uses a completely different pattern for testing (with `dgoss`). I should probably evaluate which is the better approach and then bring these into synchronization... later 😅

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,6 +49,9 @@ server-pro:  ## Build RSP image
 test-rsp: test-server-pro
 test-server-pro:
 	cd ./server-pro && IMAGE_NAME=rstudio/rstudio-server-pro:$(RSP_VERSION) docker-compose -f docker-compose.test.yml run sut
+test-rsp-i: test-server-pro-i
+test-server-pro-i:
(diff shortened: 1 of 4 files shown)


colearendt:
`timeout` research from:
https://stackoverflow.com/a/35830101/6570011
https://git.savannah.gnu.org/cgit/coreutils.git/tree/src/timeout.c

PR #42, opened 2020-04-14
Bump rspm

colearendt:
@slopp I also pointed the images at the public RSPM. Is this ok to do now?


server-pro/conf/repos.conf
```diff
@@ -1,2 +1,2 @@
+RSPM=https://packagemanager.rstudio.com/cran/__linux__/bionic/latest
```
colearendt, line 1: Changed order so binaries take precedence... I hope?

danielfrg: Good question :D

PR #43, opened 2020-04-17
Simplified and harden Dockerfile

F4stFr3ddy:
Hi,

Changes:
- Add Git to Studio Dockerfile as it is quite logical to have it present for integration with Git
- Simplify Dockerfiles, shrink size and improve readability
- Harden Dockerfiles with [Hadolint](https://github.com/hadolint/hadolint)

Let me know if you have questions / suggestions.

Best regards,

Fred

### DIFF
--- a/connect/Dockerfile
+++ b/connect/Dockerfile
@@ -9,6 +9,10 @@ ENV LANGUAGE en_US:en
 ENV LC_ALL en_US.UTF-8
 ENV DEBIAN_FRONTEND=noninteractive
 
+# Download RStudio Connect -----------------------------------------------------#
+ARG RSC_VERSION=1.8.2-10
+RUN curl -O https://s3.amazonaws.com/rstudio-connect/rstudio-connect_${RSC_VERSION}_amd64.deb
+
 # hadolint ignore=DL3008
 RUN apt-get update --fix-missing \
     && apt-get install -y --no-install-recommends \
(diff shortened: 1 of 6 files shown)


F4stFr3ddy:
Sorry about the failed tests, I blindly refactored the Connect and Package manager Dockerfile as I was mostly interested in fixing the Studio Pro one and figured might as well do the same for the other Dockerfiles.

connect/Dockerfile
```diff
@@ -9,8 +9,26 @@ ENV LANGUAGE en_US:en
 ENV LC_ALL en_US.UTF-8
 ENV DEBIAN_FRONTEND=noninteractive
 
-RUN apt-get update --fix-missing && apt-get install -y wget bzip2 ca-certificates \
-    libglib2.0-0 libxext6 libsm6 libxrender1 gdebi-core libssl1.0.0 libssl-dev git
+# Download RStudio Connect -----------------------------------------------------#
```
danielfrg, line 12: If we want to make this more optimal we should put the `curl` that downloads the product in the same `RUN` that installs and deletes the `.deb`

In general I would prefer to keep the `apt` dependencies and the `gdebi` installation still separate  `RUN` commands, as it allows us to iterate a faster when updating versions of the product eventually when the images are more stable we can optimize further but right now I prefer the iteration speed.

danielfrg:
Hi Fred, thank you very much for the PR!
Totally makes sense to add git and most of the refactors are pretty good.
I requested one change to split the installation, while its not optimal I would prefer to keep it like that for now.

connect/Dockerfile
F4stFr3ddy: Gotcha, I've moved the `gdebi` installation one of the last stages in the build to speed up the iterative build.

F4stFr3ddy:
Thanks! I've added the requested changes.

danielfrg:
Thanks for the contribution!

PR #45, opened 2020-04-23
Fix float

colearendt:
I found out why the floating container constantly needs to be reset!

- The server process has to be dead in order to deactivate the license 😑 
- We tether the container to the server process so that that can't happen

I fixed by:
- Making _sure_ that the process is dead before deactivating
- Using the `wait` command to leave the bash shell as the parent process and tether to the server process while allowing me to kill the server process if I need to (i.e. in order to deactivate)

It seems to work pretty well in testing!! Not sure how the following would be handled:
- networking blips / not able to contact the deactivation server
- if the process doesn't respond to `kill` in some context... Maybe we should `kill -9`?

This should at least make testing with the floating license server less painful.

Also, we had `PRODUCT` as an arg, but also as a required runtime env var. Yuck. Sorry about that - I fixed with `sed` substitution

### DIFF
--- a/helper/float/Dockerfile
+++ b/helper/float/Dockerfile
@@ -45,7 +45,8 @@ RUN echo '\n\
     <!-- <proxy url="http://user:pass@127.0.0.1:8080/"/> -->\n\
 </config>\n\
 \n'\
-| sed -s "s/{{PORT}}/${PORT}/" > /etc/${PRODUCT}-license-server.conf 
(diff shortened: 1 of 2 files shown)


danielfrg:
aha! Nice catch sir.

> * networking blips / not able to contact the deactivation server
Not sure about this one tbh :S

> * if the process doesn't respond to `kill` in some context... Maybe we should `kill -9`?
Yes, I think this should be fine

PR #48, opened 2020-04-28
Updating links

AshleyHenry15:
Several links needed to be updated since RSPM and RSC migrated from Bookdown to MkSocs


colearendt:
Thanks for fixing these!

PR #50, opened 2020-05-01
Fix openssl

### LINKED ISSUE #49: Problems with openssl
In the RSP image in particular

colearendt:
Close #49 

adding `/opt/python/x.x.x/bin/` to the path overwrites the system `openssl`, which can be problematic for certificate verification, updates, etc.

### DIFF
--- a/server-pro/Dockerfile
+++ b/server-pro/Dockerfile
@@ -32,7 +32,6 @@ RUN curl -o /tmp/miniconda.sh https://repo.anaconda.com/miniconda/Miniconda3-4.5
     /opt/python/${PYTHON_VERSION}/bin/conda clean -tipsy && \
     /opt/python/${PYTHON_VERSION}/bin/conda clean -a && \
     /opt/python/${PYTHON_VERSION}/bin/pip install virtualenv
-ENV PATH /opt/python/${PYTHON_VERSION}/bin:$PATH
 
 # Install other Python PyPi packages
 RUN pip install --no-cache-dir \
@@ -44,7 +43,6 @@ RUN pip install --no-cache-dir \
 
 # Add binaries to the PATH
 RUN \
-    ln -s /opt/python/${PYTHON_VERSION}/etc/profile.d/conda.sh /etc/profile.d/conda.sh && \
     ln -s /opt/python/${PYTHON_VERSION}/bin/python /usr/local/bin/python && \
     ln -s /opt/python/${PYTHON_VERSION}/bin/python3 /usr/local/bin/python3 && \
     ln -s /opt/python/${PYTHON_VERSION}/bin/rsconnect /usr/local/bin/rsconnect && \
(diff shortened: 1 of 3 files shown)


server-pro/Dockerfile
```diff
@@ -30,11 +30,8 @@ RUN curl -o /tmp/miniconda.sh https://repo.anaconda.com/miniconda/Miniconda3-4.5
...
     rm /tmp/miniconda.sh && \
     /opt/python/${PYTHON_VERSION}/bin/conda clean -tipsy && \
-    ln -s /opt/python/${PYTHON_VERSION}/etc/profile.d/conda.sh /etc/profile.d/conda.sh && \
-    echo "PATH=/opt/python/${PYTHON_VERSION}/bin:$PATH" > /etc/profile.d/python.sh && \
     /opt/python/${PYTHON_VERSION}/bin/conda clean -a && \
     /opt/python/${PYTHON_VERSION}/bin/pip install virtualenv
-ENV PATH /opt/python/${PYTHON_VERSION}/bin:$PATH
 
 # Install other Python PyPi packages
 RUN pip install --no-cache-dir \
```
danielfrg, line 37: Need to use the full `/opt/python/...` here since we are removing the `PATH` above

server-pro/Dockerfile
```diff
@@ -44,6 +41,14 @@ RUN pip install --no-cache-dir \
                 rsp_jupyter==1.2.5003 \
                 rsconnect_jupyter==1.3.3.1
 
+# Add binaries to the PATH
```
danielfrg, line 44: I think I will remove this all together, it does add this but other ones like jupyter will be missing and this list is just gonna get outdated.

server-pro/Dockerfile
```diff
@@ -44,6 +41,14 @@ RUN pip install --no-cache-dir \
...
+RUN \
+    ln -s /opt/python/${PYTHON_VERSION}/bin/python /usr/local/bin/python && \
+    ln -s /opt/python/${PYTHON_VERSION}/bin/python3 /usr/local/bin/python3 && \
+    ln -s /opt/python/${PYTHON_VERSION}/bin/rsconnect /usr/local/bin/rsconnect && \
+    ln -s /opt/python/${PYTHON_VERSION}/bin/virtualenv /usr/local/bin/virtualenv && \
+    ln -s /opt/python/${PYTHON_VERSION}/bin/conda /usr/local/bin/conda
+
 # Install Jupyter extensions
 RUN jupyter-nbextension install --sys-prefix --py rsp_jupyter && \
```
danielfrg, line 53: Same `PATH` comment as for `pip`

colearendt:
I must be doing something wrong in `launcher-env` because the `PATH` modification does not seem to be doing anything... 🤷‍♂️

danielfrg:
lgtm!

PR #58, opened 2020-05-27
Add directory init

### LINKED ISSUE #24: Explore an init / entrypoint directory
For the `postgres` docker image, they have the notion of an "init" directory that you can put scripts into to be executed. 

https://github.com/docker-library/postgres/blob/0d0485cb02e526f5a240b7740b46c35404aaf13f/12/docker-entrypoint.sh#L128

I think having such a directory (for RStudio Server Pro, at least) may be helpful for making the images more extensible. It also may be another way of starting the launcher 🤷‍♂(Related to #9)

The use case I have presently is I want to symlink the `/etc/rstudio/load-balancer` file to somewhere else (workaround for this: https://github.com/rstudio/rstudio-pro/issues/1380), and _extending_ our entrypoint script is challenging.

colearendt:
Close #24 (for RSP, at least). Probably worth pausing at RSP for a little while to see how well it works

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ See the [RStudio Server Pro guide](https://docs.rstudio.com/ide/server-pro/authe
 
 | Variable | Description | Default |
 |-----|---|---|
+| `RSP_CREATE_USER` | Whether to create a user on startup. Turn off with any value other than "true" | true |
 | `RSP_TESTUSER` | Test user to be created in the container, turn off with an empty value | `rstudio` |
 | `RSP_TESTUSER_PASSWD` | Test user password | `rstudio` |
 | `RSP_TESTUSER_UID` | Test user UID | `10000` |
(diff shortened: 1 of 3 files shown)


colearendt:
This is working as expected 😄 Mounted a few test scripts in to verify. These could be used for provisioning users, starting `sssd`, etc.

```
docker run -it --rm -e RSP_LICENSE -v $(PWD)/test.sh:/entrypoint.d/test.sh -v $(PWD)/test2.sh:/entrypoint.d/test2.sh rstudio/rstudio-server-pro:1.4.1103-4
```

colearendt:
Holding for #114 since we do not want to conflict

colearendt:
Related to #24

bschwedler:
Changes to image startup will be addressed in the in-progress image build refactor. The work is being tracked in the posit-dev GitHub org.

PR #61, opened 2020-07-16
update versions to latest

colearendt:
Update RSP and RSC versions to latest


colearendt:
Ugh. No build successes? :sob: That doesn't bode well. I think RSC changed where it installs / downloads from... for the morrow

PR #62, opened 2020-07-21
fix the connect R version installation issue

colearendt:
😑


nicolasstransky:
Thanks for the quick turnaround!

colearendt:
Our pleasure!! Apologies for the wacky issue! 😄 We added a test to hopefully protect against anything like this in the future

PR #80, opened 2020-12-23
add a simple example

### LINKED ISSUE #74: Clearly inform users wishing to use Offline Activation that this is not supported out of the box
This repo's [README.md](https://github.com/rstudio/rstudio-docker-products/blob/master/README.md) contains the following:

> **Licensing**
> Using the RStudio Server Pro Docker image requires to have a valid License. You can set the RSP license to use this in three ways:
> 1. Setting the RSP_LICENSE environment variable to a valid license key
> 2. Setting the RSP_LICENSE_SERVER environment variable to a valid license server / port
> 3. Mounting a /etc/rstudio-server/license.lic single file that contains a valid license for RStudio Server Pro

While the above implies that [Offline Activation](https://docs.rstudio.com/ide/server-pro/license-management.html#offline-activation) is not supported, this could be stated explicitly.

The above is for RSP, but the same applies to RSPM and RSC.

(See 53862.)

We might also consider adding a comment to this effect, in the Dockerfiles:

- [RSP](https://github.com/rstudio/rstudio-docker-products/blob/master/server-pro/Dockerfile)
- [RSPM](https://github.com/rstudio/rstudio-docker-products/blob/master/package-manager/Dockerfile)
- [Connect](https://github.com/rstudio/rstudio-docker-products/blob/master/connect/Dockerfile)

colearendt:
add licensing note about offline
close #74

I added the simple examples because I have seen people copy / paste the more involved examples (which will only work from the root of this repo and are designed for more "real" deployments). They invariably fail in confusing ways.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Docker images for RStudio Professional Products
 Also remember to deactivate the license before stopping the container or it will count as an active license,
 look at the `startup.sh` scripts for an example on how to do this automatically.
 
-## RStudio Server Pro
+# RStudio Server Pro
 
 #### Simple Example
 
@@ -121,7 +121,7 @@ docker run --privileged -it \
 Open [http://localhost:8787](http://localhost:8787) to access RStudio Server Pro.
 The default username and password are `rstudio`.
 
-## RStudio Connect
+# RStudio Connect
 
 #### Simple Example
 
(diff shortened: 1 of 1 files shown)


slopp (approved):
It may make sense to clarify the goal, e.g. instead of "to get started" perhaps

"To verify the basic image runs as a first step:"

Up to you though

Issue #64, opened 2020-07-31
Figure out what we want to do around "LTS"


colearendt:
The biggest problem here is for RStudio Connect. If we change R / Python versions, users who are using the image stand to experience all of their content breaking at once.

We should think through how to make that switch. Should we be providing an image that will only do _additive_ R / Python version additions for "LTS"? At least for very long periods of time? Will product changes make this easier? Actual user feedback preferred 😄

danielfrg:
It's a good point, I would prefer to keep just one docker image.

I think what we do is what you mentioned additive installs of the runtimes while also removing old (2 versions old or a specific timeline), thinking about what most languages do i think python supports only 3 minor versions at a time and with each new one they deprecate another. This usually give people a lot of time since they release around twice a year.

colearendt:
100%. That's a great call. One concern that will go along with this is image size. The images will get progressively larger 😛 

IMO we should figure out a standard, publicize it / document it, and then give a bit of burn in to implement. People will have time, but they will potentially eat updates without realizing when they upgrade. We could also coordinate "breaking" releases with "major" releases of product versions 🤷‍♂️

danielfrg:
You know I dont care much about size 🙈 

> but they will potentially eat updates without realizing when they upgrade

Agree but this happens with everything so I am not very worried about making this perfect, just good enough and standard.

> We could also coordinate "breaking" releases with "major" releases of product versions 🤷‍♂️

That could work but it might make more sense to "break it" when the runtimes stop supporting to keep it more in sync with the actual sorce.

colearendt:
Support of the containers is tied to product support

PR #88, opened 2021-01-29
DONT MERGE overall updates

slopp:
updated R
updated Python
added drivers

probably shouldn't merge this till we decide if we want to swap out R/Py versions or append them


slopp:
yikes! we also need a way to have the py version flow through to the .gcfg file ....

colearendt:
Yeah I have been debating this update, because anyone blindly using this image will immediately get broken content the next time they pull. I thought about coinciding it with an upgrade to the product version so that it would be a part of the upgrade process (but I keep forgetting to do that)... and I also punted on waiting until we have different runtimes available via the launcher 🤷‍♂️  The upgrade story is a tricky one

colearendt:
Related: https://github.com/rstudio/rstudio-docker-products/issues/64

slopp:
I think the more correct thing to do is just add a new version of R / Py

colearendt:
Closing since this is pretty old. We ended up adding R/Python versions, tracking them in NEWS.md, and trying to get them in sync across the different images. (Some old versions were replaced or removed)

PR #98, opened 2021-03-04
added postscript option to startup.sh

KVInventoR:
I would like to added option to run some postscript inside Docker image during start container.
We need to have it for configure some monitoring options inside container.


colearendt:
Thanks so much for your PR! Have you had a look at https://github.com/rstudio/rstudio-docker-products/pull/58 ? It seems to do something similar, if I am reading your PR correctly?

KVInventoR:
Hi @colearendt 
aha, yes, it looks like a similar feature
I will close my pr and wait another one when it will be merged

PR #103, opened 2021-03-24
add a dummy value for address

colearendt:
enables things like account registration and email with a placeholder
(the real solution is to accept environment variables for easy
configuration)


slopp:
I think this is fine but I'd like to see more docs; both in the readme and ideally in an in-line comment as well

PR #104, opened 2021-03-25
Add code server

colearendt:


slopp:
I can't claim to follow the sed stuff here, what is going on with the version # -> URL construction?

A couple other thoughts:

1. Docs? :) 
2. Since we are baking this into the image, do we want to install the R / Py extensions as well? IDK if the user space install makes sense in an ephemeral image?  
3. Could we take advantage of the RSP vs-code install command? The less places to maintain code server URLs... the better IMO
4. Do we need any config for RSP to actually use code server? I didn't see any changes that look like they would enable code server to be used?

colearendt:
Great questions! Yeah, the sed thing is a mess thanks to code-server shenanigans.

1. Yes! I will add that
2. I think we should just leave user-space install as the path to extensions here so we stay un-opinionated. Further, user space install will happen inside the home directory, so is typically something that has to be tackled for persistence / RSP function anyways.
3. We could do that... I opted for this pattern to follow r-session-complete and how we install there (it is a direct copy/paste). We cannot sue RSP vs-code install command for r-session-complete because the vs-code install command is not available for the session bits. Thoughts on consistency w/ r-session-complete vs. ease of maintenance?
4. Thank you. I forgot 🙈

slopp:
Copying the r-session-complete pattern makes sense for me

I can't remember - do these images serve as the basis for RSP in the helm deployment w/out local launcher as well? 

I think if this is about testing RSP quickly, it'd be nice to pre-install the extensions. But if the idea is that these images are really a basis for a production deployment in Kubernetes than you wouldn't want to install them. Overall, it is probably fine not to include them, but it is something to keep an eye on - specially once we become reliant on the RStudio plugin. 

I would recommend we include a note in the readme that explains that users should install extensions through the regular vs code interface (including the rstudio plugin) and persist them through the home directory, but that admins could extend the image to install themselves. And then in the image file perhaps we could have the commands to preinstall them included just commented out?

@MariaSemple  is the rstudio plugin that you are developing going to be installable by users through the regular VS code extension installation process through the VS Code UI?

colearendt:
Basically this entire PR is defunct now that we manage and install VSCode versions differently / using the helper. I will close this and create a new PR

PR #112, opened 2021-05-05
Name containers for easier usage and various readme updates

bdeitte:
When using this repo directly for some testing, I found it easier to have the images named so that I could more easily use docker commands against them.  It also makes for a cleaner environment, since you are creating so many images with auto-generated names that stick around.

I also changed the IMPORTANT readme section with some of my own thoughts, certainly want to make sure to talk with you about this @colearendt and can change if I am missing anything there.

I also added a section on building while being cognizant that the main use case for the README is the DockerHub readers.  I did look to see for a bit if I could split things up for DockerHub and realized this isn't super easy and likely will change as we consider different build options here.


bdeitte:
@colearendt Let me know if you are good with this and I'll merge in, thanks!

Makefile
```diff
@@ -75,7 +75,9 @@ test-server-pro-i:
 
 run-rsp: run-server-pro
 run-server-pro:  ## Run RSP container
+	docker rm rstudio-server-pro
```
colearendt, line 78: Why are you adding the `rm` here? Was there a problem you ran into? It makes sense that in some cases you want to remove the old state, but in other cases, you want the container to stick around so you can stop/start it without losing state.

(I.e. we could also add the `--rm` flag to the `docker run` command below, but that means there is no "statefulness" of an image, which I generally see as something that the user should opt into because it is destructive to state)

colearendt (commented):
This looks great, thank you!! Just a couple of questions

README.md
```diff
@@ -2,12 +2,12 @@
...
 
-**Note:** Running any RStudio profesional products inside docker requires you to have a valid license for the product.
-Also remember to deactivate the license before stopping the container or it will count as an active license,
-look at the `startup.sh` scripts for an example on how to do this automatically.
-
-**IMPORTANT:** [All images and product versions are subject to our support policy, and outdated images will be
-removed. Please make plans to update accordingly](https://rstudio.com/about/platform-support/#supported-pro-product-versions)
+**IMPORTANT:** There are a few things you need to know before using these images:
+1. Running any RStudio professional products inside docker requires you to have a valid license for the product. Also remember to deactivate the license before stopping the container or it will count as an active license- the containers will try to do this automatically but you should ensure this works in your setup.
+1. These images are not currently supported by by the RStudio support team.
```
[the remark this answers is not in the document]
bdeitte: I haven't seen the wording elsewhere, but I can ask around.  I can switch "images" to "images themselves" (and ahem, switch "by by" to "by"), would that sound more reasonable?

README.md
```diff
@@ -462,3 +462,31 @@ docker run -it --privileged \
...
+To build RStudio Package Manager:
+```
+make package-manager
+```
+
+You can alter what exactly is built by changing `server-pro/Dockerfile`, `connect/Dockerfile`, and `package-manager/Dockerfile`.
+
+You can then run what you've built to test out with the `-run` commands.  For instance, to run the server-pro container you've built:
+```
+make server-pro-run
```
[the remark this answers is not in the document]
bdeitte: Yikes, I can't believe I got all of these backwards, thanks for the fixup.

Makefile
bdeitte: If you don't add rm, the command won't work as the run always tries to create a new container.  docker start is generally used to start/stop containers, so I don't think the run command would ever use an existing container.

README.md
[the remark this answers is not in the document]
bdeitte: And updating with a few tweaks, let me know what you think!

Makefile
colearendt: Ohhhh. I see. 🙈  Yeah, that's pretty strange. What are your thoughts on whether we should:
- add `--rm` flag (which deletes on completion)
- prefix with `docker rm` (which deletes on subsequent attempts)
- use `docker-compose up service-name` to treat the container more like a service (which is idempotent and will work on subsequent attempts)

bdeitte: Personally, I like just using docker and not docker-compose, since I think people will want to see just the docker image running properly and get confused if it looks like docker-compose is needed.  I also would rather not have it deleted on completion in case people do need to do more with it.

README.md
```diff
@@ -2,12 +2,10 @@
...
-
-**IMPORTANT:** [All images and product versions are subject to our support policy, and outdated images will be
-removed. Please make plans to update accordingly](https://rstudio.com/about/platform-support/#supported-pro-product-versions)
+**IMPORTANT:** There are a few things you need to know before using these images:
+1. These images are provided as a convenience to RStudio customers and are not formally supported by RStudio. If you have questions about these images, you can ask them in the issues in the repository or to your support representative, who will route them appropriately.
+1. Outdated images will be removed periodically from DockerHub as product version updates are made.  Please make plans to update at times or use your own build of the images.
+1. These images are meant as a starting point for your needs. Consider creating a fork of this repo, where you can continue to merge in changes we make while having your own security scanning, base OS in use, or other custom changes.  We provide [instructions for building](https://github.com/rstudio/rstudio-docker-products#instructions-for-building) for these cases.
```
bdeitte, line 8: @colearendt Heads up, updated with the very latest from the internal conversations on this.

colearendt: I like it!

colearendt (approved):
LGTM!

PR #113, opened 2021-05-07
use the license-manager directly on rsp

### LINKED ISSUE #109: Handle license file consistently (rstudio-server-pro)
RStudio Server Pro handles the license file differently than the other products, and I believe this command disregards the license file location

https://github.com/rstudio/rstudio-docker-products/blob/57ab97f951d889301dd810c75234f3696149bf72/server-pro/startup.sh#L46

We should fix that!

We should also consider using a dedicated license directory, that way a Kubernetes secret volume could "automatically update." (And bonus points for consistency across products! Unfortunately a breaking change though...)

> Note: A container using a ConfigMap as a subPath volume will not receive ConfigMap updates.

From [here](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#add-configmap-data-to-a-volume)

colearendt:
stop using the rstudio-server helper script
its behavior differs from using the license-manager directly

close #109


bdeitte (approved):
So I see we are both doing other things in the current meeting, ahem.  😄   As I understand its repo better, happy to start reviewing things that come in here.  And this looks good to me- I actually didn't know there was the other way to do this, didn't notice the two different ways!

PR #118, opened 2021-05-12
update rstudio-server-pro to rstudio-workbench

colearendt:
make the .deb name customizable
trace through the customization for daily/preview builds

### DIFF
--- a/server-pro/hooks/pre_build
+++ b/server-pro/hooks/pre_build
@@ -9,6 +9,11 @@ apt-get update -qq && apt-get install -y jq
 replace_version() {
   sed -i.bak "s/^RSP_VERSION=.*/RSP_VERSION=$1/g" ./.env
 }
+
+replace_name() {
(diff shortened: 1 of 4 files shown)


server-pro/hooks/pre_build
```diff
@@ -26,6 +31,7 @@ if [[ "${DOCKER_TAG}" == *"daily"* ]]; then
     echo "Latest version found: $match"
     replace_version $match
     replace_base_url_s3
+    replace_name rstudio-workbench
```
bdeitte, line 34: I got confused by this and am curious- why replace here and not have the .env file read as this?

colearendt: Because we are in a state of flux. `preview` and `daily` need the `rstudio-workbench` name right now, where our released builds need `rstudio-sever-pro` 😭 

This is me replacing the entry in the `.env` file, which will get used by the build hook. This is another one of the "not declarative and hard to understand" things that I would like to improve in CI 😄

PR #119, opened 2021-05-12
bump connect to 1.8.8.1

colearendt:


bdeitte (approved):
A late review, but as I try to get in the habit of looking at these 👍

PR #138, opened 2021-06-14
Add r session images

colearendt:


r-session-complete/NEWS.md
```diff
@@ -0,0 +1,36 @@
+# 1.4.1106-5
+
+## Overview
+
+R is at `/opt/R/4.0.2/bin/R`
+python is at `/opt/python/3.7.7/bin/python`
+jupyter is at `/opt/python/3.7.7/bin/jupyter`
+code-server is at `/opt/code-server/code-server`
```
MariaSemple, line 8: `/opt/code-server/code-server` is deprecated and will be removed in newer versions, but the bin path is there on older versions.

```suggestion
code-server is at `/opt/code-server/bin/code-server`
```

MariaSemple (changes requested):

.github/workflows/build-latest.yaml
```diff
@@ -46,116 +49,46 @@ jobs:
...
+        if: ${{ github.ref == 'refs/heads/main' }}
         run: |
-          docker push rstudio/${{ matrix.config.product }}:${{ matrix.config.version }}
-          docker push rstudio/${{ matrix.config.product }}:latest
-          docker push ghcr.io/rstudio/${{ matrix.config.product }}:${{ matrix.config.version }}
-          docker push ghcr.io/rstudio/${{ matrix.config.product }}:latest
+          docker push rstudio/${{ matrix.config.product }}:${{ matrix.config.tag_prefix }}${{ matrix.config.version }}
+          docker push rstudio/${{ matrix.config.product }}:${{ matrix.config.tag_prefix }}latest
+          docker push ghcr.io/rstudio/${{ matrix.config.product }}:${{ matrix.config.tag_prefix }}${{ matrix.config.version }}
+          docker push ghcr.io/rstudio/${{ matrix.config.product }}:${{ matrix.config.tag_prefix }}latest
```
bdeitte, line 94: Can you explain what all of these changes do?  I didn't follow all of it from one read through at least.  And is this anything that @nahara7 needs to know about with her upcoming testing work?

r-session-complete/README.md
```diff
@@ -0,0 +1,59 @@
+# docker-r-session-complete
+
+Dockerfiles for R and Python sessions and jobs to be used RStudio Workbench,
+Launcher, and Kubernetes.
```
bdeitte, line 4: When this is merged in I would recommend we change all of the docker-r-session-complete repo README to say the repo is deprecated and point here, to reduce confusion.

bdeitte: Also are you solving the DockerHub README issue as part of this work?  DockerHub takes the root README always, right?

.github/workflows/build-latest.yaml
colearendt: This is a no-op for all of the existing images, because `tag_prefix` resolves to nothing. For the `r-session-complete` images, we specify `tag_prefix=bionic-` or `tag_prefix=centos7-`so that the images will become `bionic-latest` and `centos7-latest`, for instance, because the r-session-complete images have two OS's.

This is mostly done for backwards compatibility with the existing setup. It shouldn't affect her testing at all 😄

r-session-complete/README.md
colearendt: Yes, definitely will update the docker-r-session-complete repo to point here. 

As for the DockerHub README, the current doc improvement plan is:
- update / decouple the docs here in the repo
- copy those to DockerHub

There are ways to automate this with GHA, but it's not immediately clear how well they will work with our setup. Planning to take a couple of cycles improving the docs once the meat and potatoes have landed 😄

bdeitte: Makes sense, sounds like a good plan!

colearendt:
@MariaSemple It looks like all of the `/opt/code-server/code-server` references were in the NEWS. I added a new NEWS entry for 1.4.1717-3 to update, but I think it's ok to leave the history in the NEWS file for reference.

MariaSemple:
Ah, I thought I saw a reference in vscode.conf.

PR #139, opened 2021-06-15
workbench - add R 4.1.0 and python 3.9.3

colearendt:
I also considered swapping `r-base:3.6` for `r-base:4.0` - thoughts?

Feel free to merge if you want the workbench image to be updated! (Now is a good time, because we just swapped the image from rstudio-server-pro to rstudio-workbench)


slopp:
Thanks Cole! I am really excited to see the multiple versions of R showcased here, and I think folks who pull this image out of the box will benefit from the new version of R.

A couple of things to perhaps discuss / add before we merge:

- I'd like to see the goss tests for RSW updated to check for these 2 versions of R
- Should we add compatible alt versions of R to the RSC image? And tests?

server-pro/NEWS.md
```diff
@@ -0,0 +1,4 @@
+# 1.4.1717-3
+
+- Added a version of R. Currently including 4.1.0, 3.6.3
+- Currently includes Python version 3.6.5
```
bdeitte, line 4: ❤️

bdeitte (approved):
Not sure if adding this to connect should be a separate ticket, included in here, or not done.  But as-is this looks good to me.

colearendt:
@ian-flores What do you think of the python changes? I registered 3.9.3 as a kernel, but following the pattern from https://github.com/rstudio/sol-eng-demo-server, did not register `/opt/python/jupyter/bin/python` as a kernel (to keep them "separate"). This means that we still only have one python "kernel" in Jupyter, but we do have two pythons installed and the pattern in the Dockerfile is easy to copy / paste

bdeitte:
Perhaps for a separate PR- but since the title of this PR says "workbench" I should note the README still says "RStudio Server Pro" 😄

server-pro/Dockerfile
```diff
@@ -32,29 +32,52 @@ RUN apt-get update --fix-missing && apt-get install -y --no-install-recommends \
...
+ARG PYTHON_VERSION=3.9.5
+RUN curl -O https://repo.anaconda.com/miniconda/Miniconda3-4.7.12.1-Linux-x86_64.sh && \
+    bash Miniconda3-4.7.12.1-Linux-x86_64.sh -bp /opt/python/${PYTHON_VERSION} && \
+    /opt/python/${PYTHON_VERSION}/bin/conda install -y python==${PYTHON_VERSION} && \
+    /opt/python/${PYTHON_VERSION}/bin/pip install \
+        ipykernel \
+        virtualenv \
+        && \
+    rm -rf Miniconda3-*-Linux-x86_64.sh && \
+    /opt/python/${PYTHON_VERSION}/bin/python -m ipykernel install --name py${PYTHON_VERSION} --display-name "Python ${PYTHON_VERSION}"
```
bdeitte, line 80: I remember seeing a NEWS file in one of your recent PRs.  Could we have all of this in one as well?  I really like the idea of starting to put changes like this in a NEWS/CHANGES file.

server-pro/NEWS.md
```diff
@@ -0,0 +1,12 @@
...
+- Added R version 4.1.0. Image includes versions 4.1.0, 3.6.3
+- BREAKING: change image name to `rstudio/rstudio-workbench`  
+- BREAKING: change the location / version of python. Now includes:
+  - version 3.8.10 at /opt/python/jupyter to run jupyter
+  - version 3.9.5 at /opt/python/3.9.5 as a kernel for jupyter
+
+# 1.4.1106-5
+
+- Includes R version 3.6.3 at /opt/R/3.6.3
+- Includes Python version 3.6.5 at /opt/python/3.6.5
```
bdeitte, line 12: Thanks for adding another one, great to see these.

bdeitte:
Realizing now I reviewed before, but took another looks and still looks good to me.

colearendt:
Yeah, we definitely need to take a pass at replacing all `server pro` stuff with `workbench`. My plan was to include that in more of the "docs" work once we have functionality where we want it.

ian-flores:
@colearendt I took a look at the files updated. I think it looks good. However, I still think we should add two python kernels to Jupyter to showcase a similar functionality to the one displayed by the RStudio IDE. I also believe we should register an R kernel to showcase full interoperability between IDEs and languages but I know there has been pushback of this in the past.

colearendt:
@ian-flores Done. I think one thing worth noting - adding the second python kernel increased image size from 3.6 GB to 4.3 GB. 

It's also worth noting that our current image size is 2.5GB, so this PR does not quite double image size, but comes close.

I think registering an R kernel should be a separate discussion / PR.

Issue #14, opened 2019-12-20
Harden images


colearendt:
In general follow https://docs.docker.com/develop/develop-images/dockerfile_best-practices/

delete the apt-cache after each apt-install, delete downloaded files after extracting or running

colearendt:
I also think it'd be nice to test the launcher functions with `verify-installation` for RSP 🤷‍♂️

JosiahParry:
Giving this a +3 vote from DOD (three groups have requested this that I can recall)
Particularly for Centos7. DOD uses openshift for launcher

PR #142, opened 2021-06-21
Dockerfiles Security Best Practices

nahara7:
# Intent
Update Dockerfiles to include more security best practices.

Addresses: #14 

# Approach

- Added a pipefail setting in connect 

- cleaned apt-cache in package manager


connect/Dockerfile
```diff
@@ -27,7 +27,7 @@ RUN apt-get update --fix-missing \
     && rm -rf /var/lib/apt/lists/*
 
 # Install Python  -------------------------------------------------------------#
-ARG PYTHON_VERSION=3.6.5
+ARG PYTHON_VERSION=3.9.5
```
bdeitte, line 30: I assume this will be fine, but I have no idea if Connect has any recommended Python versions.  If no one says "yes this looks ok" I'll track someone down to ask.

connect/Dockerfile
```diff
@@ -49,6 +49,7 @@ RUN chmod +x /usr/local/bin/startup.sh
 
 # Download RStudio Connect -----------------------------------------------------#
 ARG RSC_VERSION=1.8.8.2
+SHELL [ "/bin/bash", "-o", "pipefail", "-c"]
```
bdeitte, line 52: Is this something we do in other Dockerfiles, or is this a best practice you see somewhere?  Just curious on this as I wasn't sure about this.

nahara7: Yes ! I found it here: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#using-pipes .

connect/Dockerfile
colearendt: So the one downside here is that anyone using this docker image will immediately have all of their python content break when they update to use this image. It might be good to hold this update until we bump the Connect version to 1.9.0, so that we coincide it with a major Connect version change. 

We also have Connect's launcher capabilities coming which will improve Connect's flexibility here. For reference, we are having similar discussion about changing R versions here: https://github.com/rstudio/rstudio-docker-products/pull/139

connect/Dockerfile
colearendt: Nice! TIL about the `SHELL` command: https://docs.docker.com/engine/reference/builder/#shell

bdeitte (approved):
Looks good to me to merge

colearendt (approved):
Same!

PR #144, opened 2021-06-23
add content examination script alongside image definition

aronatkins:


colearendt (approved):
Looks great to me!! Is there a way to easily generate the list of images that are enumerated in the github action? Or did you do that manually?

aronatkins:
The example commands were manually crafted. Between that list in the README, the GitHub action matrix, and the `build-images.sh`, we now have three shapes of the same thing. I wish that weren't the case but don't have great ideas for improvement.

colearendt:
Yeah, me either. I think one idea for generation would be to enumerate public images using the docker hub or github API with some type of filter. I think that can wait though. This looks good to me! Shall we merge?

PR #147, opened 2021-06-26
Add r session debug images

colearendt:


r-session-debug/bionic/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+FROM rstudio/r-session-complete:bionic-${RSP_VERSION}
+
+RUN apt-get update -qq \
+    && apt-get install -y \
+        dnsutils \
+        tcpdump \
+        netcat \
+        net-tools \
+        psutils \
+        vim \
```
colearendt, line 11: Should also have strace if it is not already installed

PR #149, opened 2021-06-30
add a daily cron

colearendt:
for preview builds

It's possible that we may want to set up a webhook so that Connect (or other products) can tell us when a new build is available / has landed on master 🤷‍♂️ 

https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_dispatch

### DIFF
--- a/.github/workflows/build-preview.yaml
+++ b/.github/workflows/build-preview.yaml
@@ -9,6 +9,12 @@ on:
   pull_request:
     branches:
       - main
+  workflow_dispatch:
+    inputs:
+      which:
+        description: "Which images to build. Default 'all'. Options: 'all','rstudio-workbench-preview','rstudio-connect-preview','rstudio-connect-content-init-preview','r-session-complete-preview'"
+        required: true
+        default: "all"
 
 name: build/test/push (preview)
 jobs:
@@ -16,6 +22,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     name: build-${{ matrix.config.tag_prefix}}${{ matrix.config.product}}
+    if: ${{ github.events.inputs.which == 'all' || github.events.inputs.which == matrix.config.product }}
 
     strategy:
       fail-fast: false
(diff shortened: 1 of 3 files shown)


.github/workflows/build-preview.yaml
```diff
@@ -1,4 +1,8 @@
 on:
+  schedule:
```
aronatkins, line 2: How do you feel about workflow dispatch to let folks manually trigger? https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_dispatch

```yaml
on:
  workflow_dispatch:
    ...
```

aronatkins (approved):

.github/workflows/build-preview.yaml
colearendt: I am totally up for that 😄 Just wasn't sure what parameters / etc. we should set up

aronatkins: Oh, right. This is against all products because of the matrix. I was thinking that we would be able to trigger the build of a single image, possibly also overriding the version from "latest". I'm not sure how easy it is to accomplish since all of the builds are driven by this one action configuration.

colearendt: So fortunate and unfortunate news. It is pretty easy to set up a webhook input to trigger only a single image build. I have not customized version yet, but that wouldn't be too bad either to have another input. Right now it just grabs the latest from that json blob that we output from CI. 

On the downside, enabling `webhook_dispatch` as a trigger seems to turn off the other triggers (i.e. build on PRs, main, cron, etc.). I'm going to reach out to GitHub / GItHub community to sanity check if that is the case, but for now I just copied the workflow into a duplicate file. We already have one set of duplicates (basically - between latest and preview), so why not three!? haha.

colearendt:
@aronatkins does this still look good to you? If so I'll go ahead and merge and we can start having a cron / playing with the webhook!

.github/workflows/build-preview.yaml
colearendt: Now that I think about it, if we decide to do the version override, I think I would add a `--override` flag to the `get_version.py` script so that it is easy to inject into the `get-version` step and keep the workflows more or less identical

aronatkins (approved):

PR #151, opened 2021-07-02
Add content-base-pro image builds

dbkegley:
Adds a `build-pro` make target. All pro images are built `FROM` the existing content-base images for each distro (bionic, right now obviously)

### DIFF
--- a/content-base/bionic/pro/Dockerfile
+++ b/content-base/bionic/pro/Dockerfile
@@ -1,0 +1,21 @@
+ARG BASE_IMAGE
+
+FROM ${BASE_IMAGE}
+
+MAINTAINER RStudio Docker <docker@rstudio.com>
+
+# Install RStudio Professional Drivers ----------------------------------------#
+ARG DRIVERS_VERSION
+ARG R_VERSION
+
+RUN apt-get update -y && \
+    DEBIAN_FRONTEND=noninteractive apt-get install -y unixodbc unixodbc-dev gdebi && \
+    rm -rf /var/lib/apt/lists/*
+
+RUN curl -O https://drivers.rstudio.org/7C152C12/installer/rstudio-drivers_${DRIVERS_VERSION}_amd64.deb && \
+    apt-get update && \
+    DEBIAN_FRONTEND=noninteractive gdebi --non-interactive rstudio-drivers_${DRIVERS_VERSION}_amd64.deb && \
+    rm -rf /var/lib/apt/lists/* && \
+    cp /opt/rstudio-drivers/odbcinst.ini.sample /etc/odbcinst.ini
+
+RUN /opt/R/${R_VERSION}/bin/R -e 'install.packages("odbc", repos="https://packagemanager.rstudio.com/cran/__linux__/bionic/latest")'
--- a/content-base/Makefile
+++ b/content-base/Makefile
@@ -15,6 +15,7 @@ DISTRIBUTION=bionic
 TAG_BASE=rstudio/content-base
 R_VERSION=3.6.3
 PYTHON_VERSION=3.7.6
+DRIVERS_VERSION=1.7.0
 
 all: push
 .PHONY: all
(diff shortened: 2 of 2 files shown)


content-base/bionic/pro/Dockerfile
```diff
@@ -0,0 +1,21 @@
+ARG BASE_IMAGE
+
+FROM ${BASE_IMAGE}
+
+MAINTAINER RStudio Docker <docker@rstudio.com>
+
+# Install RStudio Professional Drivers ----------------------------------------#
+ARG DRIVERS_VERSION
+ARG R_VERSION
```
aronatkins, line 9: It's unfortunate that we're not automatically getting R_VERSION from the parent `Dockerfile`.

If `R_VERSION` were an environment variable, it would be exposed from parent to child. Not sure if that's the preferred option, here.

The distribution (bionic) is also inferred from the base image (used in the RSPM URL).

I don't have any real suggestions, here.

aronatkins (commented):
Other references to these images:

* The `content-base/README.md`: https://github.com/rstudio/rstudio-docker-products/blob/main/content-base/README.md
* The `content-base/build-images.sh` script: https://github.com/rstudio/rstudio-docker-products/blob/main/content-base/build-images.sh
* GH CI: https://github.com/rstudio/rstudio-docker-products/blob/main/.github/workflows/build-content-base.yaml

@colearendt - how would you prefer that we organize this second image? A separate top-level directory? The approach here?

content-base/bionic/pro/Dockerfile
colearendt: Top of the head suggestion so I don't forget when I review this next week - what about using `find /opt/R` to execute this installation for every version of R in the directory, thereby rendering the `R_VERSION` arg unnecessary

content-base/bionic/pro/Dockerfile
```diff
@@ -0,0 +1,21 @@
+ARG BASE_IMAGE
```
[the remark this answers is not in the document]
colearendt: The one strange thing about building this FROM the other images is that it uses the ones that were built on a previous pass. 

I like the idea of `bionic-pro/Dockerfile` as a path, but don't have any great ideas for resolving the dependency tree other than "document and ignore" or "build in some type of delay / dependency / etc."

I guess the other option would be to completely copy the whole Dockerfile 😢 We could always copy the github action and have it run as a dependency after the other one completes

colearendt (approved):
I think I'm leaning towards creating another copy of the GitHub action to trigger this, and make it a dependency of the other content images.

I think `bionic-pro` is probably a better path for things.

colearendt (changes requested):
Oops. Wrong status 🙈

dbkegley:
Just to make sure I understood this feedback correctly:

- Update the base path for the new pro images to be `content-base/bionic-pro`
- Add a new github action that is a dependency of the `content-base/bionic` build

My intent for building FROM the existing content-base images was to avoid all the duplication in the dockerfile and to use the build cache to speed up the build. Is there an alternative approach for these pro images that's preferred?

colearendt:
That sounds good to me! The other thing we could do is create a path called `content-pro/bionic`. 

I like the pattern of using the FROM to reduce duplication, etc. It's just a bit tricky to set up the builds properly for that. I think the pattern you describe is perfect and will get that to happen properly 😄 I'm not sure how to optimize that type of dependency tree - on one hand, for testing, we could probably run concurrently, but for building and publishing images, we definitely want them to be a dependency 🙈

content-base-pro/Makefile
```diff
@@ -0,0 +1,46 @@
...
+# See: https://github.com/rstudio/r-builds/blob/0eaf4ff908f5944c152f89128bbffb3f73ce1c95/Makefile#L7-L8
+build:
+	docker build --build-arg R_VERSION=$(R_VERSION) \
+	--build-arg PYTHON_VERSION=$(PYTHON_VERSION) \
+	--build-arg DRIVERS_VERSION=$(DRIVERS_VERSION) \
+	--build-arg BASE_IMAGE="$(TAG_BASE):r$(R_VERSION)-py$(PYTHON_VERSION)-$(DISTRIBUTION)" \
+	-t $(TAG_BASE_PRO):r$(R_VERSION)-py$(PYTHON_VERSION)-$(DISTRIBUTION) $(DISTRIBUTION)
+.PHONY: build
+
+# Do not run push if you are not Cole.
```
colearendt, line 38: We can remove this. 😄  And honestly, now that we have CI, maybe we should remove the manual push? Or at least make it not the default?

We can also follow-up this PR with another one to make the change to `content-base` as well, if that's preferable / easier.

content-base-pro/README.md
```diff
@@ -0,0 +1,15 @@
...
+these pro images. Attempting to build a pro image using a content-base image that
+does not exist will result in an error.
+
+The make targets for building these pro images are the same as the the targets for the base images.
+
+
+## Github Actions
+
+Because of the dependency on the content-base images, the github actions that build these images
+depend on completion of the base image builds in [build-content-base](../.github/workflows/build-content-base.yaml)
```
colearendt, line 15: Should we rename the yaml file?

colearendt (approved):
This looks awesome! I'm honestly happy with how this looks. Just a few ideas, but not anything necessary

We could also think about refactoring our folder structure a bit...? It might make it easier to colocate things like the matrix.json and the README / etc.

- content
    - base
    - pro

instead of:

- content-base
- content-base-pro
?

I guess this forces the question: should we have `content-base` and `content-pro` images? Or `content-base` and `content-base-pro`?

dbkegley:
I think I agree on all points. How does this sound?

- `content-base` and `content-pro` for the image names
- reorg the folder structure (and rename `content` to `connect-content`):
  - connect-content
     - base/bionic
     - pro/bionic
- rename `.github/build-content-base.yaml` to `.github/build-connect-content.yaml`

colearendt:
I like that a lot!! With the one small caveat that I lean towards stripping `connect` out, since the hope (long term) is that these images could be used to back content in the IDE (i.e. if we can mount in the IDE bits), or outside of Connect too!

colearendt (approved):
Looks awesome @dbkegley ! Well done!! 😄

Issue #116, opened 2021-05-10
Add a way to provide an environment variable for `--verify-installation` command on RSP


colearendt:
RStudio Server has a helpful `--verify-installation` that is very hard to use in docker. It would be nice if our entrypoint pivoted on an environment variable or something to know that it is in "diagnose" mode and runs the `--verify-installation` to a provided directory (ideally a shared directory...)

PR #153, opened 2021-07-07
Verify-Installation Command

nahara7:
Addresses: #116 

## Intent
To simplify use of verify-installation command inside the container.

## Approach
Verify Installation command was tricky to use in docker. This runs the command at startup and writes its output to a path referenced by the ` DIAGNOSTIC_DIR ` environment variable. Allows image tests to reference the log file. Additional environment variables added include:

`DIAGNOSTIC_ENABLE` : false by default. Runs verify installation command at startup

`DIAGNOSTIC_ONLY` : false by default. Only runs verify-installation command and then exits the script.


server-pro/Dockerfile
```diff
@@ -108,6 +108,7 @@ ENV RSP_TESTUSER_PASSWD rstudio
 ENV RSP_TESTUSER_UID 10000
 ENV RSP_LAUNCHER true
 ENV RSP_LAUNCHER_TIMEOUT 10
+ENV DIAGNOSE_DIR /var/log
```
colearendt, line 111: Let's do this to match with the logging spec https://app.tettra.co/teams/rstudio/pages/rstudio-logging-standard
```suggestion
ENV DIAGNOSE_DIR /var/log/rstudio
```

server-pro/startup.sh
```diff
@@ -68,6 +68,15 @@ if [ "$RSP_LAUNCHER" == "true" ]; then
   wait-for-it.sh localhost:5559 -t $RSP_LAUNCHER_TIMEOUT
 fi
 
+#verify installation
+if [ ! -z "$DIAGNOSE_DIR" ]; then
+   echo "==VERIFY INSTALLATION==";
+   mkdir -p $DIAGNOSE_DIR/verify-installation
+   chown rstudio-server:rstudio-server $DIAGNOSE_DIR/verify-installation
```
colearendt, line 75: Instead of chowning, let's set the directory permissions to `chmod 777` so that the directory is writable 😄  (again, consistent with the logging spec)

server-pro/startup.sh
```diff
@@ -68,6 +68,15 @@ if [ "$RSP_LAUNCHER" == "true" ]; then
   wait-for-it.sh localhost:5559 -t $RSP_LAUNCHER_TIMEOUT
 fi
 
+#verify installation
+if [ ! -z "$DIAGNOSE_DIR" ]; then
+   echo "==VERIFY INSTALLATION==";
+   mkdir -p $DIAGNOSE_DIR/verify-installation
```
colearendt, line 74: Let's just use the `$DIAGNOSTIC_DIR` directly instead of creating a dedicated subfolder

server-pro/startup.sh
```diff
@@ -68,6 +68,15 @@ if [ "$RSP_LAUNCHER" == "true" ]; then
   wait-for-it.sh localhost:5559 -t $RSP_LAUNCHER_TIMEOUT
 fi
 
+#verify installation
+if [ ! -z "$DIAGNOSE_DIR" ]; then
```
colearendt, line 72: Maybe `DIAGNOSTIC_DIR` instead of `DIAGNOSE_DIR`?

colearendt (changes requested):
This looks great! A couple of random ideas 😄

colearendt:
Another idea I just had - when this "diagnostic report" ends, it will start the product as normal. Maybe we should have an environment variable that, if set, causes the script to exit after the diagnostic report ends? i.e. `DIAGNOSTIC_ONLY` or something like that.

server-pro/startup.sh
colearendt: Also, if we are going to set a default, then we unfortunately cannot use the existence of this env var as a toggle 😢 

I think it is workable to have `DIAGNOSTIC_ENABLE` or some type of flag variable that turns the execution on 😄

nahara7:
> Another idea I just had - when this "diagnostic report" ends, it will start the product as normal. Maybe we should have an environment variable that, if set, causes the script to `exit 0` after the diagnostic report ends? i.e. `DIAGNOSTIC_ONLY` or something like that.

I like this ! Will do !

server-pro/startup.sh
```diff
@@ -24,6 +24,15 @@ deactivate() {
 }
 trap deactivate EXIT
 
+# Call verify installation command and write to log file
+verify_installation () {
+   echo "==VERIFY INSTALLATION==";
+   mkdir -p $DIAGNOSTIC_DIR
+   chmod -R 777 $DIAGNOSTIC_DIR
```
colearendt, line 31: Let's not use `-R` here - that's a bit destructive, potentially 😄 Just dropping the `-R` should be sufficient!
```suggestion
   chmod 777 $DIAGNOSTIC_DIR
```

server-pro/startup.sh
```diff
@@ -24,6 +24,15 @@ deactivate() {
...
+   su rstudio-server -c "touch ${DIAGNOSTIC_DIR}/verify.log"
```
colearendt, line 32: We also shouldn't need to initialize this since we are not tailing the log file elsewhere - can remove 😄

server-pro/startup.sh
```diff
@@ -68,6 +77,16 @@ if [ "$RSP_LAUNCHER" == "true" ]; then
   wait-for-it.sh localhost:5559 -t $RSP_LAUNCHER_TIMEOUT
 fi
 
+# Check diagnostic configurations
```
colearendt, line 80: I think a nested if might be a bit clearer here. Something like:
```
if [ "$DIAGNOSTIC_ENABLE" == "true" ]; then
  verify_installation
  if [ "$DIAGNOSTIC_ONLY" == "true" ]; then
    echo "Exiting script because DIAGNOSTIC_ONLY=${DIAGNOSTIC_ONLY}"
    exit 0
  fi;
else
  echo "not running verify installation because DIAGNOSTIC_ENABLE=${DIAGNOSTIC_ENABLE}"
fi
```

Also, it might be good to ensure that `verify_installation` is on the PATH? (I'm not sure).

colearendt (commented):
Looking really good! A few more little tidbits 😄

server-pro/startup.sh
```diff
@@ -68,6 +68,19 @@ if [ "$RSP_LAUNCHER" == "true" ]; then
...
+if [ "$DIAGNOSTIC_ENABLE" == "true" ]; then
+  /usr/local/bin/verify_installation.sh &
+  sleep 7
```
colearendt, line 74: I'm curious - why the sleep?

server-pro/Dockerfile
```diff
@@ -108,6 +108,13 @@ ENV RSP_TESTUSER_PASSWD rstudio
 ENV RSP_TESTUSER_UID 10000
 ENV RSP_LAUNCHER true
 ENV RSP_LAUNCHER_TIMEOUT 10
+ENV DIAGNOSTIC_DIR /var/log/rstudio
+ENV DIAGNOSTIC_ENABLE true
```
colearendt, line 112: Let's set this to false by default. Having every server startup running diagnostic seems a little extreme by default 😄

server-pro/startup.sh
```diff
@@ -68,6 +68,19 @@ if [ "$RSP_LAUNCHER" == "true" ]; then
   wait-for-it.sh localhost:5559 -t $RSP_LAUNCHER_TIMEOUT
 fi
 
+# Check diagnostic configurations
+if [ "$DIAGNOSTIC_ENABLE" == "true" ]; then
+  /usr/local/bin/verify_installation.sh &
```
colearendt, line 73: Ahh interesting approach - what led you to backgrounding this process / separating into a separate script?

That would definitely make running it on every startup more palatable. It does make the interleaving log output a bit trickier though.

server-pro/Dockerfile
nahara7: Okay !

server-pro/startup.sh
nahara7: I originally had a wait command. It took quite some time to fetch the job output at the end of the line.
`...Job has finished running
Fetching job output...`
The sleep command allowed me to view the file when I was sure the directory existed but not waiting as long. Is there a different way you think I should do ?

server-pro/startup.sh
nahara7: I only did this because I saw your comment about ensuring if verify_installation was on the path ! I can put it back to being a function within the startup.sh. 
It does intertwine with the logs ! Especially when it shows the job output.

colearendt: Ohhh sorry. Yeah, what I meant was that before you had `verify-installation` (IIRC). I just meant that I wasn't familiar with a shell script called `verify-installation`, and if there was one, be sure that invoking it that way works 😄  In the shell script, you used what I was more familiar with `rstudio-server verify-installation` 😄 

I do think a function in the script is probably easier for now. And running it in blocking fashion should be good. Then we don't have to worry about waiting and whatnot 😄 

The goal here (which I probably did not articulate very clearly - my apologies!), is to have a flag that users can opt into that will give them the diagnostics output when they are actively involved in changing / setting up / troubleshooting / etc. In most cases, we want this to no-op, IMO 😄

nahara7: I see ! I put the function back into the startup script. Though at the `DIAGNOSTIC_ONLY` condition, I did use a sleep command to display the log output before exiting the script

nahara7: I will take it out of the background ! It was freezing for me when I did that though, or there was a lot of lag. ^^

colearendt: Ah interesting! Yeah, that's probably just the verify-installation command running, I would hope/guess 😮 So long as `tee` does streaming output... which I believe it does IIRC.

If this command is being run, it is the whole / main reason that the container exists, so hopefully the time to execute shouldn't cause concern in users.

nahara7: Okay makes sense. I was accounting for the instance where `DIAGNOSE_ENABLE` =true but `DIAGNOSE_ONLY` =false. Here it freezes until you exit the program, but the users would want the product started normally.

colearendt: Ohhh weird... it does? I wonder why... That definitely shouldn't be the case 😮

bdeitte (approved):
Looks good to me!  My only suggestion would be that we document the environment variables somewhere in a README.  But my guess is this is a more wide-ranging issue related to docs and not sure if it conflicts with any of @colearendt's doc work that I haven't gotten a chance to look through.

colearendt:
Ok, had a long chat w/ the RSW team today. TL;DR; I think we need to:
- remove the `&` / backgrounding of the `verify-installation` command
- remove the `sleep 5`

What was happening with the hang you experienced was the product was stuck in a state where it could not start the job, but wasn't erroring and also wasn't timing out. The team is now aware and will hopefully improve the product to be a bit more clear in the long run, but backgrounding the process can actually hide some of these concerning issues.

This is an example that reproduces the problem (using this branch / the public image after we merge):
```
docker run -v $(pwd)/server-pro/startup.sh:/usr/local/bin/startup.sh -e DIAGNOSTIC_ENABLE=true -e DIAGNOSTIC_ONLY=true -it --rm rstudio/rstudio-workbench:1.4.1717-3
```

And this example works as we expect (added `--privileged`) (I was mounting the startup script so I did not have to rebuild the image with each tweak)
```
docker run --privileged -v $(pwd)/server-pro/startup.sh:/usr/local/bin/startup.sh -e DIAGNOSTIC_ENABLE=true -e DIAGNOSTIC_ONLY=true -it --rm rstudio/rstudio-workbench:1.4.1717-3
```

I'm not super opinionated about docs, to be honest. I am AOK if we merge this and then follow up with docs once things have stabilized in the repo a bit. We've done a lot of work in this PR already 😄 

I'll send y'all a link to the slack thread privately

colearendt (approved):
Woohoo! Looks awesome!

Issue #154, opened 2021-07-07
Workbench logging improvements


colearendt:
RStudio Workbench currently has a very complicated `startup.sh` / entrypoint script and logging structure that:

- writes logs to several directories (i.e. `/var/log/`, `/var/lib/rstudio-server/monitor`, `/var/lib/rstudio-launcher`, etc.)
- tails these logs to stdout (which interleaves output together in potentially confusing ways, but only works for files that are initialized with proper permissions by the startup script itself)
- tails the logs again at container shutdown (potentially duplicating messages) to ensure that no log messages are missed (if the container dies before another "tail" polll)

In daily builds for RStudio Workbench (see `rstudio/rstudio-workbench-preview` or `ghcr.io/rstudio/rstudio-workbench-preview` images), we purportedly have some logging improvements! We should:

- Edit `logging.conf` so that all logging output (ideally for workbench AND launcher) comes to stdout / stderr
- Ensure that no log files are being written to disk
- Ensure that log level changes (i.e. info -> debug) maintain stdout/stderr and no files being written to disk
- Change the `startup.sh` script to remove all of the "log file tailing, initializing, etc." stuff

A few notes:
- We should check on the status of "workbench running launcher as a subprocess" - so we do not have to start the process ourselves
- It is possible to start the container interactively, mount config files, and then run the different services manually so that iteration is quicker

### Acceptance Criteria
- [ ] RSW container prints all logs to stdout and stderr
- [ ] RSW container does not use any of the default file logging locations
- [ ] When the RSW log level is changed, no files are used

colearendt:
1 - In discussion with @kevinhankens , we were hopeful that the launcher and RSW logging changes have landed. I'm not sure whether RSW is managing the launcher process or not yet, though
2 - Yes! https://hub.docker.com/r/rstudio/rstudio-workbench-preview or https://github.com/orgs/rstudio/packages/container/package/rstudio-workbench-preview
3 - I haven't thought about this a whole lot. We could pivot startup.sh based on the type of build in the GitHub Actions yaml, or we can get the branch working and then wait for the release of the next version to merge 😄

kevinhankens:
Awesome, thanks! For 1. It sounds like the launcher logs might be dependent on https://github.com/rstudio/rstudio-pro/pull/2713 but I'm not 100% sure.

bdeitte:
[...]

PR #157, opened 2021-07-13
WorkBench Logging Improvements

nahara7:
Improvements in RSW allow for global settings of logs to go to stderr. This branch incorporates these changes by editing the logging.conf and simplifying the startup.sh.

Addresses: #154 

After updating the logging.conf file, stdout and stderr of workbench and launcher go to stderr. The following files no longer exist:
 `/var/log/rstudio-server.log`
 `/var/lib/launcher.log`,
`/var/lib/rstudio-launcher/Local/rstudio-local-launcher*.log`,
 `/var/lib/rstudio-launcher/Kubernetes/rstudio-kubernetes-launcher.log`, 
`/var/lib/rstudio-server/monitor/log/rstudio-server.log`

This can be verified by running the container as privileged and examining the processes for workbench and launcher within the container via `/proc/{PID}/fd`. Both their file descriptors 1 and 2 are piped to the same number. i.e `pipe:[NUM]` and their logs are shown on the screen.

Log outputting to stderr, with none of the previous files existing,  is consistent with both `info` and `debug` log levels.


colearendt (approved):
This looks awesome! Well done!

I created a `dev` branch that builds dev / preview images so that we can start to try some of this stuff before release

PR #158, opened 2021-07-14
add simple MIT license

colearendt:
I'm not sure whether I should clarify anywhere in `LICENSE.md` what this license applies to? It is in the _repository_ and it says "this software," so I think it should inherently apply to the code / docs in the repository and nothing else, correct?

For reference: [licensing images themselves is complex](
https://www.linuxfoundation.org/resources/publications/docker-containers-what-are-the-open-source-licensing-considerations/) because it includes the licenses of every piece of software included in _any layer_ in the image.


server-pro/Dockerfile
```diff
@@ -123,7 +123,8 @@ RUN apt-get update --fix-missing \
     && gdebi --non-interactive ${RSP_NAME}-${RSP_VERSION}-amd64.deb \
     && rm ${RSP_NAME}-${RSP_VERSION}-amd64.deb \
     && apt-get clean \
-    && rm -rf /var/lib/apt/lists/*
+    && rm -rf /var/lib/apt/lists/* \
+    && rm -rf /var/lib/rstudio-server/r-versions
```
colearendt, line 127: Woops - this was supposed to go in a different PR 🙈

akgold (approved):
My comment from the other repo here too. I think it's known that "this software" refers to the code actually in the repo.

Good with me, assuming you're removing that code before merging.

Issue #155, opened 2021-07-07
Add functional-service testing for goss


colearendt:
It would be nice to test that services actually run successfully / properly in goss. However, this is hard because goss tests are parallelized. 

We need to add licenses into secrets for GitHub Actions and make easy make targets / etc. for testing locally (i.e. `- RSC_LICENSE` in `docker-compose.test.yml` will use an env var that is defined in the user's environment variables without needing to put them in a `.env` file / etc.). You can look at how licenses are handled here if that's helpful: https://github.com/rstudio/sol-eng-docker#getting-started 

A couple of possible patterns:

1. Run the product inside of "timeout" so that we get an expected output if the product runs successfully for X seconds
https://github.com/rstudio/rstudio-docker-products/blob/de5f9c3cdbf03307d500f51225772f756c6e1aa9/connect/test/goss.yaml#L35
    - If we use this pattern, we should probably test / document the different behaviors
    - Downside: quite confusing setup, not using the startup script (in this example)

1. Run the product interactively:
    - i.e. https://github.com/nahara7/rstudio-docker-products/blob/08fd9e03e0456d2908ac213f37d1ab5979e06f32/package-manager/test/goss.yaml#L45
    - Downside: if the product is successful, then it will not return / the goss test will timeout
    - Also, we should look into why this is exiting 0 when "unhappy" (i.e. the product failed)
    - Downside: not using the startup script (in this example)

1. Run the product as a part of the testing script, and then check for listening in goss:
    - We could probably use `tini -- /usr/local/bin/startup.sh` consistently because that is the CMD/ENTRYPOINT for all of the product images. We will need to run into the background, maybe write to a log file, and then cat the log file if the tests exit unhappily? Getting logs is a secondary matter to testing that the product functions 😄 
    - could inject into https://github.com/rstudio/rstudio-docker-products/blob/main/server-pro/test/run_tests.sh
    - Downside: we are (potentially) more limited in what we can test in the goss.yaml because the service will already be running. This should be mostly academic / not a real downside though.

I lean towards option 3. What are your thoughts!?

A couple of notes:

- we will likely need to add the `privileged: true` flag to `docker-compose.test.yml`
- we will need to take a look at GitHub Secrets and see how they behave across forks, etc. @nahara7 if you'd like to work on this repo so that we can get "true" CI, then we can add your access and use a branch here rather than your fork!
- I would recommend testing some of these things interactively so you get a sense for what the goss tests are doing
[...]

PR #164, opened 2021-07-19
Functional Service Testing

nahara7:
### Intent
To provide Goss tests for Connect, Package Manager, and Workbench to ensure they are running and that their configuration files (`rstudio-connect.gcfg`, `rstudio-pm.gcfg`, `rserver.conf`) are of the correct format

Addresses #155 

## Approach
Start up service in the beginning of each corresponding `run_tests.sh` and write its output to a temporary `startup.log` file added in the dockerfile. 
1. Additional Goss  test to see if the process is running 
2. Additional Goss test to check that the `startup.log` file contains no errors from trying to read the service's configuration file.
### Secrets
Secret licenses for each image were added into github actions. These are retrieved in the Github workflows and added to each `docker-compose` file.


Makefile
```diff
@@ -109,7 +108,7 @@ test-connect-i:
 
 run-rsc: run-connect
 run-connect:  ## Run RSC container
-	docker rm rstudio-connect
```
colearendt, line 112: @bdeitte added these (and discussed their reasoning) here: https://github.com/rstudio/rstudio-docker-products/pull/112

Generally speaking, it's good to proceed with caution modifying too many things (even if you think they are unnecessary) in one go, since it can provide friction for reviewers to review the change that you care more about. In more formal projects especially, you want to ensure that the only changes you make are related to the issue you are working on, and create a separate issue if you want to discuss / get buy in on tangential / other related changes.

(That said, if you need such changes to get your dev work working smoothly, this is something that I have done before - modifying locally and just not committing the changes since they are personal preferences. And if that gets frustrating "enough," then its probably worth opening an issue to discuss improvements 😄 )

connect/Dockerfile
```diff
@@ -39,13 +39,17 @@ RUN curl -o /tmp/miniconda.sh https://repo.anaconda.com/miniconda/Miniconda3-4.5
...
 ARG TINI_VERSION=0.18.0
 RUN curl -L -o /usr/local/bin/tini https://github.com/krallin/tini/releases/download/v${TINI_VERSION}/tini && \
     chmod +x /usr/local/bin/tini
 
+
+
 COPY startup.sh /usr/local/bin/startup.sh
-RUN chmod +x /usr/local/bin/startup.sh
+RUN chmod +x /usr/local/bin/startup.sh && \
+     touch /tmp/startup.log
```
colearendt, line 52: Is there a reason for the touch statements? I think we can remove these `touch` statements for a few reasons:
1. these logs are used only in testing / are not essential to the proper functioning of the service, so they should not be baked into the images
2. the reason we had the "touch" pattern before was so that our `tail -f` commands could find them. In general practice, touching files is definitely something to avoid unless there is a particular reason for doing so 😄

connect/startup.sh
```diff
@@ -19,7 +19,7 @@ elif test -f "/etc/rstudio-connect/license.lic"; then
     /opt/rstudio-connect/bin/license-manager activate-file /etc/rstudio-connect/license.lic
 fi
 
-# lest this be inherited by child processes
+# let this be inherited by child processes
```
colearendt, line 22:
```suggestion
# ensure these cannot be inherited by child processes
```

`lest` was actually intentional. Me and my nerdy [old english](https://www.merriam-webster.com/dictionary/lest) 😅

connect/test/run_tests.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+# start connect
+/opt/rstudio-connect/bin/connect --config /etc/rstudio-connect/rstudio-connect.gcfg 2>/tmp/startup.log
```
colearendt, line 4: Let's start Connect the same way the container does `tini -- /usr/local/bin/startup.sh`. That should handle licensing better too.

connect/test/run_tests.sh
```diff
@@ -18,3 +21,9 @@ curl -sL https://github.com/aelsabbahy/goss/releases/download/v$GOSS_VERSION/gos
   && GOSS=/tmp/goss
 
 GOSS_FILE=$GOSS_FILE GOSS_VARS=$GOSS_VARS $GOSS v --format documentation --max-concurrent $GOSS_MAX_CONCURRENT
+
```
colearendt, line 24: Drop unnecessary whitespace, please! 😄 VSCode may have a way to do this automatically.

colearendt (changes requested):
Looking awesome!! A couple of feedback items that apply to all of the different products 😄 I didn't go through and tag each instance of it - thought that might get a little busy 🙈

connect/test/run_tests.sh
colearendt: Also, I think you need to background these processes, or they will block the execution of the rest of the service.

connect/Dockerfile
nahara7: Okay ! I had originally mounted them as temporary files. Would you suggest going back to that ?

connect/test/run_tests.sh
```diff
@@ -1,6 +1,10 @@
 #!/bin/bash
 
-# install goss
+# create temporary startup log file
+cat > /tmp/startup.log
```
colearendt, line 4: Is this necessary? 😮

connect/test/run_tests.sh
```diff
@@ -1,6 +1,10 @@
...
+
+# start connect
+tini -- /usr/local/bin/startup.sh 2>/tmp/startup.log &
```
colearendt, line 7:
```suggestion
tini -- /usr/local/bin/startup.sh > /tmp/startup.log 2>&1 &
```

Otherwise `stdout` will still go to the terminal 😄

server-pro/docker-compose.test.yml
```diff
@@ -13,6 +13,8 @@ services:
       - PYTHON_VERSION_ALT
       - JUPYTER_PYTHON_VERSION
       - R_VERSION_ALT
+      - RSW_LICENSE
```
colearendt, line 16: This RSW_LICENSE will not _actually_ work until we switch RSP to RSW. Should we switch back to RSP for now, and take the RSW bit as a part of follow-up? Alternatively, we might could do something like this to fill the gap:

```suggestion
      - RSP_LICENSE=${RSW_LICENSE}
```

colearendt (commented):
This is looking really good! Just a couple of items / ideas (which apply to all 3 places)

connect/test/run_tests.sh
nahara7: Is there another way you think we should create the log file ?

colearendt: It should be created automatically by the lines below 😄 We only have to create the log file _before starting the process_ if we need to `tail` the file. Hopefully we are moving away from that need across the board 😄

nahara7: I see !

connect/startup.sh
```diff
@@ -6,7 +6,7 @@ set -x
 # Deactivate license when it exists
 deactivate() {
     echo "Deactivating license ..."
-    /opt/rstudio-connect/bin/license-manager deactivate >/dev/null 2>&1
+    /opt/rstudio-connect/bin/license-manager deactivate >/dev/null 2>&1 &
```
colearendt, line 9: Why is this backgrounded?

server-pro/test/run_tests.sh
```diff
@@ -18,3 +19,4 @@ curl -sL https://github.com/aelsabbahy/goss/releases/download/v$GOSS_VERSION/gos
   && GOSS=/tmp/goss
 
 GOSS_FILE=$GOSS_FILE GOSS_VARS=$GOSS_VARS $GOSS v --format documentation --max-concurrent $GOSS_MAX_CONCURRENT
+
```
colearendt, line 22: Missing a trailing newline here

workbench/conf/launcher.pem
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
colearendt, line 1: We definitely don't want these files committed 😄 Also, we should probably add them to .gitignore if they aren't already?

colearendt (commented):
Looking awesome! Just a few little nits and then we should be good to go!! Well done!

connect/test/run_tests.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 
-# install goss
+# start connect
+tini -- /usr/local/bin/startup.sh 2>/tmp/startup.log 2>&1 &
```
colearendt, line 4:
```suggestion
tini -- /usr/local/bin/startup.sh >/tmp/startup.log 2>&1 &
```

package-manager/test/run_tests.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 
-# install goss
+# start package manager
+tini -- /usr/local/bin/startup.sh 2>/tmp/startup.log 2>&1 &
```
colearendt, line 4:
```suggestion
tini -- /usr/local/bin/startup.sh >/tmp/startup.log 2>&1 &
```

server-pro/test/run_tests.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 
-# install goss
+# start rstudio-server
+rstudio-server start 2>/tmp/startup.log 2>&1 &
```
colearendt, line 4:
```suggestion
tini -- /usr/local/bin/startup.sh >/tmp/startup.log 2>&1 &
```

colearendt (approved):
One last little round of tweaks, and then I think we are good to go!! Well done!

connect/test/run_tests.sh
colearendt: `2>` says that `stderr` should go to this file. Then `2>&1` says `stderr` should go to the place where `stdout` went.

Instead, we want: `stdout` goes to file, and `stderr` goes to where `stdout` went ( so everything goes to the file ) 😄

colearendt:
Also, don't worry about the CI failures. Those are failing on main too - the preview builds have changed how they are notated, so as long as it is preview Workbench / the r-session-complete, failure is expected ATM

server-pro/test/run_tests.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 
-# install goss
+# start rstudio-server
+rstudio-server start >/tmp/startup.log 2>&1 &
```
colearendt, line 4: Why did you go back to this one? If we invoke with `tini -- /usr/local/bin/startup.sh`, we behave like the image does 😄

nahara7: I was about to ask you ! I was just testing because I noticed the `pgrep rserver` test fails when we startup with `tini -- /usr/local/bin/startup.sh` but it works when using `rstudio-server start`. Do you have any insights on why ?

colearendt: Woops - sorry for the late reply here. I am suspicious that something in the product is failing. Licensing issue or otherwise. If you are doing this interactively (there are docs on how to do that here: https://github.com/rstudio/rstudio-docker-products/blob/main/RELEASE.md#testing-locally - I have not used these in a while, so hopefully they still work!), then take a look at `/tmp/startup.log` to see what the problem is, maybe?

PR #178, opened 2021-07-30
Add common license file path

jonyoder:
Adds a common license file path `/etc/rstudio-license/license.lic` for all products. Support fallback to old path.

### DIFF
--- a/connect/startup.sh
+++ b/connect/startup.sh
@@ -11,13 +11,13 @@ deactivate() {
 trap deactivate EXIT
 
 # Activate License
-CONNECT_LICENSE_FILE_PATH=${CONNECT_LICENSE_FILE_PATH:-/etc/rstudio-connect/license.lic}
+RSC_LICENSE_FILE_PATH=${RSC_LICENSE_FILE_PATH:-/etc/rstudio-connect/license.lic}
 if ! [ -z "$RSC_LICENSE" ]; then
     /opt/rstudio-connect/bin/license-manager activate $RSC_LICENSE
 elif ! [ -z "$RSC_LICENSE_SERVER" ]; then
     /opt/rstudio-connect/bin/license-manager license-server $RSC_LICENSE_SERVER
-elif test -f "$CONNECT_LICENSE_FILE_PATH"; then
-    /opt/rstudio-connect/bin/license-manager activate-file $CONNECT_LICENSE_FILE_PATH
+elif test -f "$RSC_LICENSE_FILE_PATH"; then
+    /opt/rstudio-connect/bin/license-manager activate-file $RSC_LICENSE_FILE_PATH
 fi
 
 # lest this be inherited by child processes
--- a/connect/startup.sh
+++ b/connect/startup.sh
@@ -15,6 +15,8 @@ if ! [ -z "$RSC_LICENSE" ]; then
     /opt/rstudio-connect/bin/license-manager activate $RSC_LICENSE
 elif ! [ -z "$RSC_LICENSE_SERVER" ]; then
     /opt/rstudio-connect/bin/license-manager license-server $RSC_LICENSE_SERVER
+elif test -f "/etc/rstudio-licensing/license.lic"; then
+    /opt/rstudio-connect/bin/license-manager activate-file /etc/rstudio-licensing/license.lic
 elif test -f "/etc/rstudio-connect/license.lic"; then
     /opt/rstudio-connect/bin/license-manager activate-file /etc/rstudio-connect/license.lic
 fi
--- a/connect/startup.sh
+++ b/connect/startup.sh
@@ -11,14 +11,13 @@ deactivate() {
 trap deactivate EXIT
 
 # Activate License
+CONNECT_LICENSE_FILE_PATH=${CONNECT_LICENSE_FILE_PATH:-/etc/rstudio-connect/license.lic}
 if ! [ -z "$RSC_LICENSE" ]; then
     /opt/rstudio-connect/bin/license-manager activate $RSC_LICENSE
 elif ! [ -z "$RSC_LICENSE_SERVER" ]; then
     /opt/rstudio-connect/bin/license-manager license-server $RSC_LICENSE_SERVER
-elif test -f "/etc/rstudio-licensing/license.lic"; then
-    /opt/rstudio-connect/bin/license-manager activate-file /etc/rstudio-licensing/license.lic
-elif test -f "/etc/rstudio-connect/license.lic"; then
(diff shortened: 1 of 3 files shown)


connect/startup.sh
```diff
@@ -15,6 +15,8 @@ if ! [ -z "$RSC_LICENSE" ]; then
     /opt/rstudio-connect/bin/license-manager activate $RSC_LICENSE
 elif ! [ -z "$RSC_LICENSE_SERVER" ]; then
     /opt/rstudio-connect/bin/license-manager license-server $RSC_LICENSE_SERVER
+elif test -f "/etc/rstudio-licensing/license.lic"; then
```
colearendt, line 18: I like this approach, but what if we used the pattern of having an env var that defines the path? `RSTUDIO_LICENSE_FILE_PATH` or some such? Then just set the default to what it currently is for backwards compatibility, and override if needed.

For consistency, we would probably need to prefix it with the product: `RSC_LICENSE_FILE_PATH`, etc.

I'm torn on which is preferable though

colearendt: Honestly, yes - the case that you have one secret with licenses from each of the products in a directory. Then you might want to mount it at `/mnt/rstudio-licensing/rspm-file.lic`, `/mnt/rstudio-licensing/rsc-file.lic`, and `/mnt/rstudio-licensing/rsw-file.lic` or something

jonyoder: I like the simplicity of a hard-code path, and I don't love adding env vars. That said, I think it depends a bit on how we expect customers to use this:

1. If we expect customers with deeply-customized environments to still use this docker image, then the env var would be nice.
2. If we expect such customers to create their own custom Dockerfiles, then it's really not necessary to support that option here.

colearendt: Yeah, I agree. I think the trick is that our _licensing_ is something that I would expect our customers to copy and use, regardless of whether they are building their own image or not.

In a perfect world, this entire entrypoint would be unnecessary and the products themselves (or the license-manager itself) would be smart enough to start / read the environment properly / etc.

So it might be better for us to think about what pattern we would want in a "docker friendly license manager" and do that. Would our license manager require that you have to put the license file in a particular place? Would it take an env var input or config value to direct it?

jonyoder:
@colearendt I tried modifying this to use an env var with a default instead. It seems to work fine.

connect/startup.sh
```diff
@@ -11,12 +11,13 @@ deactivate() {
...
 
 # Activate License
+CONNECT_LICENSE_FILE_PATH=${CONNECT_LICENSE_FILE_PATH:-/etc/rstudio-connect/license.lic}
 if ! [ -z "$RSC_LICENSE" ]; then
     /opt/rstudio-connect/bin/license-manager activate $RSC_LICENSE
 elif ! [ -z "$RSC_LICENSE_SERVER" ]; then
     /opt/rstudio-connect/bin/license-manager license-server $RSC_LICENSE_SERVER
-elif test -f "/etc/rstudio-connect/license.lic"; then
-    /opt/rstudio-connect/bin/license-manager activate-file /etc/rstudio-connect/license.lic
+elif test -f "$CONNECT_LICENSE_FILE_PATH"; then
```
colearendt, line 19:
```suggestion
elif test -f "$RSC_LICENSE_FILE_PATH"; then
```

connect/startup.sh
```diff
@@ -11,12 +11,13 @@ deactivate() {
...
+    /opt/rstudio-connect/bin/license-manager activate-file $CONNECT_LICENSE_FILE_PATH
```
colearendt, line 20:
```suggestion
    /opt/rstudio-connect/bin/license-manager activate-file $RSC_LICENSE_FILE_PATH
```

connect/startup.sh
```diff
@@ -11,12 +11,13 @@ deactivate() {
 trap deactivate EXIT
 
 # Activate License
+CONNECT_LICENSE_FILE_PATH=${CONNECT_LICENSE_FILE_PATH:-/etc/rstudio-connect/license.lic}
```
colearendt, line 14:
```suggestion
RSC_LICENSE_FILE_PATH=${RSC_LICENSE_FILE_PATH:-/etc/rstudio-connect/license.lic}
```

The other place we can set this default (the other place that we set it) is in the docker image build itself. I kinda like this approach though

server-pro/startup.sh
```diff
@@ -45,12 +45,13 @@ chown rstudio-server:rstudio-server /var/lib/rstudio-launcher/Kubernetes
 su rstudio-server -c 'touch /var/lib/rstudio-launcher/Kubernetes/rstudio-kubernetes-launcher.log'
 
 # Activate License
+RSTUDIO_LICENSE_FILE_PATH=${RSTUDIO_LICENSE_FILE_PATH:-/etc/rstudio-server/license.lic}
```
colearendt, line 48:
```suggestion
RSW_LICENSE_FILE_PATH=${RSW_LICENSE_FILE_PATH:-/etc/rstudio-server/license.lic}
```
The `RSW` prefix is in flight and won't land for a little while, but that's the current plan for the future

server-pro/startup.sh
```diff
@@ -45,12 +45,13 @@ chown rstudio-server:rstudio-server /var/lib/rstudio-launcher/Kubernetes
...
 if ! [ -z "$RSP_LICENSE" ]; then
     /usr/lib/rstudio-server/bin/license-manager activate $RSP_LICENSE
 elif ! [ -z "$RSP_LICENSE_SERVER" ]; then
     /usr/lib/rstudio-server/bin/license-manager license-server $RSP_LICENSE_SERVER
-elif test -f "/etc/rstudio-server/license.lic"; then
-    /usr/lib/rstudio-server/bin/license-manager activate-file /etc/rstudio-server/license.lic
+elif test -f "$RSTUDIO_LICENSE_FILE_PATH"; then
```
colearendt, line 53:
```suggestion
elif test -f "$RSW_LICENSE_FILE_PATH"; then
```

server-pro/startup.sh
```diff
@@ -45,12 +45,13 @@ chown rstudio-server:rstudio-server /var/lib/rstudio-launcher/Kubernetes
...
+    /usr/lib/rstudio-server/bin/license-manager activate-file $RSTUDIO_LICENSE_FILE_PATH
```
colearendt, line 54:
```suggestion
    /usr/lib/rstudio-server/bin/license-manager activate-file $RSW_LICENSE_FILE_PATH
```

colearendt (approved):
Just a few prefix changes for consistency! 

I'm not a huge fan of this approach in that it is not elegant / is harder to reason about, but i do like giving the flexibility to customers 😄

PR #176, opened 2021-07-29
RSW Rebranding

nahara7:
# Intent
To rebrand `Rstudio Server Pro` to `Rstudio Workbench`

Addresses: #160 

## Updates To 

-  CI Paths
-  Makefile
- docker-compose.yaml
- RSP prefixes
- server-pro directory name
- NEWS.md

Issue #160, opened 2021-07-14
Rename RSP to RSW


colearendt:
The RStudio Server Pro product has been rebranded to RStudio Workbench. As a result, we should:

- change the `server-pro` directory to `workbench`
- update all CI paths to function
- change `RSP_` prefixed env vars to `RSW_` prefixes
- Add `NEWS.md` entries to note the breaking change (this one is going to be a bit pernicious. Helm charts will be affected, users will be affected on next container pull, etc.). Maybe we should see how close the RSW preview release is to see if we can coordinate
- Trace through docs and update any `server-pro` / `Server Pro` / etc. naming
- Update `Makefile` targets and related docs
- Update `docker-compose.yaml` examples and naming

colearendt:
Closed in #211 / #176 . Will be finally merged in #216 , but we can call it done for now.

PR #179, opened 2021-08-02
Support RSP_ or RSW_ Prefixes for licensing

jonyoder:
We recently added the var `RSW_LICENSE_FILE_PATH`, but the existing licensing env vars are prefixed with `RSP_`. This updates the startup script to support `RSW_` as an alternate prefix to improve support for our helm work.


colearendt (approved):
Nice! I like this as a good intermediate state before #160 and https://github.com/rstudio/rstudio-docker-products/pull/176

PR #188, opened 2021-08-19
Bumping drivers from 1.7 to 1.8

nwstephens:
Bumping drivers from 1.7 to 1.8


colearendt:
Closing in favor of #201 because forks are hard

PR #196, opened 2021-08-31
Bump rsc version

colearendt:
Depends on #195 landing first

- update version to first calver release (2021.08.0)
- change directory to start capturing and mapping two keys rather than three (1.9.0 -> 2021.08).
- Also fix a bug where we were only capturing one digit in each version number (i.e. would have broken on 1.9.10)


aronatkins (approved):
We should declare, somewhere, that RSC_VERSION can only ever be a YYYY.MM.PATCH version; older versions do not work with this Dockerfile.

aronatkins:
For the future-us: Here's a `sed` invocation that rewrites any-number of version components into N-1 components.

```bash
sed -r -e 's/(([0-9]+\.)+[0-9]+)\.[0-9]+/\1/'
```

aronatkins:
More future-us notes: Here are a couple `awk` invocations:

```bash
# should be all-awk
awk -F. 'BEGIN{OFS="."} { for (i=1; i<NF; i++) printf "%s%s", $i, (i<NF-1 ? OFS : ORS)}'
# only works with recent awk
awk 'NF{NF-=1}1' FS='.' OFS='.'
```

Issue #206, opened 2021-09-17
Issue with Floating Licensing (RSW)


colearendt:
We should check our docker image! There is a report that floating licensing is not working properly 😢

colearendt:
Have not been able to reproduce this. It is possible that missing this line in config was the problem: https://github.com/rstudio/rstudio-docker-products/blob/main/server-pro/conf/rserver-float.conf

PR #207, opened 2021-09-21
Fix float

colearendt:
Attempting to address #206 , but could not reproduce issues

Some doc updates may be in order, though!


bschwedler:
Outdated. The float images are now on Ubuntu 22.04

PR #208, opened 2021-09-21
r-session-complete jupyter symlink

### LINKED ISSUE #197: r-session-complete jupyter symlink
Could it be possible to add a symlink from /usr/local/bin/jupyter to /opt/python/PYTHON_VERSION/bin/jupyter so that the rstudio workbench default configurations work with the r-session-complate image without further configurations?

https://docs.rstudio.com/ide/server-pro/jupyter-sessions.html

samssann:
fixes #197


colearendt:
Hey @samssann ! Thanks for the PR! This looks great! A couple of things:

- would you mind batching this command into the previous command with `&& \` so that we don't create another layer needlessly? It's also nice coupling related things together so that these commands are easily COPY/PASTE-able
- have you built and tested this behavior locally? Sometimes symlinks can do funny things 😄 
- if you have tested locally, does jupyter lab work as well? I presume it would - I am just not familiar with how the reference from `jupyter lab` works.

colearendt:
Closed in favor of #297 (with the same outcome). Thanks!

PR #217, opened 2021-09-30
Fix for url / encoding

AndrewMcClain:
Build system does not yet encode + with a -, instead have to %2b
Switched to rstudio-ide-builds bucket to support dailies as well as
released

### DIFF
--- a/workbench/Dockerfile
+++ b/workbench/Dockerfile
@@ -123,7 +123,7 @@ RUN chmod +x /usr/local/bin/startup.sh
 
 # Install RStudio Workbench --------------------------------------------------#
 ARG RSW_VERSION=2021.09.0+351.pro6
-ARG RSW_DOWNLOAD_URL=https://s3.amazonaws.com/rstudio-ide-build/server/bionic/amd64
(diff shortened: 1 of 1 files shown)


workbench/Dockerfile
```diff
@@ -123,10 +123,10 @@ RUN chmod +x /usr/local/bin/startup.sh
 
 # Install RStudio Workbench --------------------------------------------------#
 ARG RSW_VERSION=2021.09.0+351.pro6
-ARG RSW_DOWNLOAD_URL=https://download2.rstudio.org/server/bionic/amd64
+ARG RSW_DOWNLOAD_URL=https://s3.amazonaws.com/rstudio-ide-build/server/bionic/amd64
```
colearendt, line 126:
```suggestion
ARG RSW_DOWNLOAD_URL=https://download2.rstudio.org/server/bionic/amd64
```

colearendt (changes requested):
We are using s3 / ide-builds, IIRC (we set it in the workflows file). I think the key thing will be to use the `%2b` for now... we probably need to do the same in the `r-session-complete` builds for now though.

colearendt (approved):
NVM! I didn't realize that we never switched those 🙈

PR #218, opened 2021-10-04
remove data science packages from session-complete

edavidaja:
- **BREAKING**: R and Python packages are no longer installed by default in session images.

We have removed R and Python packages from the session image.  This should:
- reduce the image build time and size
- make it easier for data scientists to develop reproducible environments
- eliminate conflicts between system libraries and subsequent user updates

If the removal of these system libraries introduces some problems for you, please comment or file an issue.

**Also:** Fixes CI and the tests for R and Python versions that upgrade with this release as well

### DIFF
--- a/r-session-complete/README.md
+++ b/r-session-complete/README.md
@@ -52,10 +52,8 @@ These images include the following layers:
 * RSW session components
 * System packages required for R, R packages, and RStudio Professional Drivers
 * One version of R
-* R packages (devtools, tidyverse, shiny, and rmarkdown)
 * One version of Python
 * Jupyter Notebooks, JupyterLab, and RSW/RSC notebook extensions
-* Python packages
 * RStudio Professional Drivers
 
 # Licensing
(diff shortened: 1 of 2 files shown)


colearendt:
Looks like the 3.7.7 check may never have been run because of this issue 🙈 I think switching `stdout` to `stderr` works sometimes? Different python versions behave differently 😑 

```
Command: /opt/python/3.7.7/bin/python --version: stdout: patterns not found: [3.7.7]
```

edavidaja:
It actually looks like jupyter and python don't get found at all?

r-session-complete/NEWS.md
```diff
@@ -3,6 +3,14 @@
...
 - **BREAKING**: Change `R_VERSION` to 4.1.0
 - **BREAKING**: Change `PYTHON_VERSION` to 3.9.5
+- **BREAKING**: R and Python packages are no longer installed by default in session images.
+
+We have removed R and Python packages from the session image.  This should:
+- reduce the image build time and size
+- make it easier for data scientists to develop reproducible environments
+- eliminate conflicts between system libraries and subsequent user updates
+
+If the removal of these system libraries introduces some problems for you, please file an issue.
```
bdeitte, line 13: The README will need updates as well

PR #231, opened 2021-11-16
Bring Connect and Content images into parity with common sys-dependencies installed by setup-assistant 

sagerb:
Connect Setup Assistant (as well as the manual instructions) provide a list of common system dependencies required by the "common" list of R packages (top 80%). This list (as found / maintained in `https://github.com/rstudio/connect-setup/blob/main/scripts/sysrec/r-pkgs`) is used to query RSPM to determine a list of system dependencies to install.

I have taken the generated list of system dependencies for Ubuntu 18.04, and added any missing ones to the docker files driving the connect image and the content base image.

The connect image was pretty bare. With the changes, the image size has increased from 2.94GB to 3.97GB.

The content base image was fairly complete and was only missing a few dependencies. The size in GB did not appear to change for these image sizes.

closes rstudio/connect#20122

### DIFF
--- a/connect/Dockerfile
+++ b/connect/Dockerfile
@@ -14,7 +14,6 @@ RUN apt-get update --fix-missing \
     && apt-get install -y --no-install-recommends \
         bzip2 \
         ca-certificates \
-        default-jdk \
         gdal-bin \
         gdebi-core \
         git \
--- a/connect/Dockerfile
+++ b/connect/Dockerfile
@@ -12,17 +12,54 @@ ENV DEBIAN_FRONTEND=noninteractive
 # hadolint ignore=DL3008,DL3009
 RUN apt-get update --fix-missing \
     && apt-get install -y --no-install-recommends \
-        wget \
         bzip2 \
         ca-certificates \
-        libglib2.0-0 \
-        libxext6 \
-        libsm6 \
-        libxrender1 \
+        default-jdk \
+        gdal-bin \
         gdebi-core \
+        git \
+        gsfonts \
+        imagemagick \
+        libcairo2-dev \
+        libcurl4-openssl-dev \
+        libfontconfig1-dev \
+        libfreetype6-dev \
+        libfribidi-dev \
+        libgdal-dev \
+        libgeos-dev \
+        libgl1-mesa-dev \
+        libglib2.0-0 \
+        libglpk-dev \
+        libglu1-mesa-dev \
+        libgmp3-dev \
+        libharfbuzz-dev \
+        libicu-dev \
+        libjpeg-dev \
+        libmagick++-dev \
+        libmysqlclient-dev \
+        libpng-dev \
+        libproj-dev \
+        libsm6 \
+        libsodium-dev \
+        libssh2-1-dev \
+        libssl-dev \
         libssl1.0.0 \
-        libssl-dev git \
-	sudo \
+        libtiff-dev \
+        libudunits2-dev \
+        libv8-dev \
+        libxext6 \
+        libxml2-dev \
+        libxrender1 \
+        make \
+        perl \
+	    sudo \
+        tcl \
+        tk \
+        tk-dev \
+        tk-table \
+        unixodbc-dev \
+        wget \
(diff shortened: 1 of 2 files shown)


connect/Dockerfile
```diff
@@ -12,17 +12,54 @@ ENV DEBIAN_FRONTEND=noninteractive
...
-	sudo \
+        libtiff-dev \
+        libudunits2-dev \
+        libv8-dev \
+        libxext6 \
+        libxml2-dev \
+        libxrender1 \
+        make \
+        perl \
+	    sudo \
```
colearendt, line 55: why is this indented differently?

connect/Dockerfile
```diff
@@ -12,17 +12,54 @@ ENV DEBIAN_FRONTEND=noninteractive
 # hadolint ignore=DL3008,DL3009
 RUN apt-get update --fix-missing \
     && apt-get install -y --no-install-recommends \
-        wget \
         bzip2 \
         ca-certificates \
+        default-jdk \
```
colearendt, line 17: I forget if the `default-jdk` causes trouble with R or if we should block down to an older version 🙈 We can either merge and see what happens or do a bit more testing

colearendt (commented):
LGTM! Just a few nits

connect/Dockerfile
aronatkins: @colearendt To confirm: You're OK expanding the default Connect image with these dependencies. It's not required for our ongoing work to expand the content images.

This expansion may also push us to have separate Connect images (for when execution is external to this image).

michaelmayer2: > I forget if the `default-jdk` causes trouble with R or if we should block down to an older version 🙈 We can either merge and see what happens or do a bit more testing

Maybe worth considering running a R CMD javareconf for R 4.1.0 and the default R version in the docker image as well to ensure rJava and other packages that depend on it can be installed.

colearendt: Yeah @aronatkins, I think the benefit from increased image-usefulness is worth the cost, personally. What data we have from customers suggest they would rather not worry about sysdeps 🙈 Those who have weighed in from SE seem to agree.

I agree that we are probably marching towards the road of having a "minimal" image down the road too. The same could be said for Workbench probably, too, which I suspect is also a bit overkill for launching sessions via the launcher.

colearendt: Now that I think about it more, let's drop the JDK and create an issue to track it across the other images too. We should decide on a consistent approach across RSW, RSW sessions, and RSC. (If RSC session image already has it we can either leave or remove 🤷‍♂️ ). 

The one place we are using this (Colorado / our demo environment), it's [not even working](https://github.com/rstudio/sol-eng-demo-server/issues/29). So probably worth tracking this as a follow-up and not holding this PR up with debates on the topic 😄  Future us can deal with it

sagerb: @cole, The inclusion of the `default-jdk` has been included in the system dependencies installed from the Setup Assistant for all Ubuntu targets. The common package list we use contains `rJava` (which is then what causes this dependency to be called out and installed. 
 Is this a problem we just haven't heard about from the field as well?

sagerb: And should we remove `rJava` from our most common list?

sagerb: After talking with Aron and understanding better what Cole was saying.. I'm going to remove the JDK from dependencies and I'll create a new issue to track this as suggested. The key here was that the feeling is that the version of jdk to be installed would probably not match what was needed.

colearendt: Yeah, spot on. I think container vs. server is very different b/c on a server you can just go add what you want if needed. And customers who need rJava are likely both rare-ish and opinionated.

sagerb:
Updates made, ready for re-review. @colearendt What can we do about the CI failure which appears to still be an issue?

colearendt (approved):
LGTM!

CI will be resolved soon-ish when RSW releases again (which should fix the version number they have posted / available)

PR #234, opened 2021-11-16
Additional R / Python content images

sagerb:
Adding additional content pro images to supply these combinations of R and Python:

- R:4.0.5, Python:3.7.10
- R:4.0.5, Python:3.6.13
- R:3.6.3, Python:3.6.13
- R:3.6.3, Python:2.7.18
- R:3.5.3, Python:2.7.18
- R:3.4.4, Python:3.7.10
- R:3.3.3, Python:3.6.13

closes rstudio/connect#20124

Initial verification will be done w/ successful CI.


content/matrix.json
```diff
@@ -8,5 +8,12 @@
   {"r": "4.0.5", "py": "3.8.8", "drivers": "1.8.0", "os": "bionic"},
   {"r": "4.0.5", "py": "3.9.2", "drivers": "1.8.0", "os": "bionic"},
   {"r": "4.1.0", "py": "3.8.8", "drivers": "1.8.0", "os": "bionic"},
-  {"r": "4.1.0", "py": "3.9.2", "drivers": "1.8.0", "os": "bionic"}
+  {"r": "4.1.0", "py": "3.9.2", "drivers": "1.8.0", "os": "bionic"},
```
aronatkins, line 11: nit: the existing entries were sorted (R version as primary sort); could we do the same?

aronatkins (approved):

connect/NEWS.md
```diff
@@ -1,9 +1,11 @@
-# 2021.10.0
+# 2021.11.0
```
aronatkins, line 1: @colearendt - should we retain the 2021.10.0 version since 2021.11.0 is not yet public?

aronatkins (commented):
Waiting for feedback from @colearendt before approval / merge

connect/NEWS.md
```diff
@@ -1,9 +1,11 @@
...
 
 - Add a bunch of system dependencies. This makes the image build bigger (~4GB),
   but ensures that R and Python content / packages are much more likely to
   build and run without error
 
+- Add additional content images.
```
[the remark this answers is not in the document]
colearendt: I wouldn't put it here since it's not related to the connect image. My opinion would be either a `NEWS.md` in the `content` directory, or else we don't bother 🤷‍♂️  😅

connect/NEWS.md
colearendt: Yeah, I agree. The general idea is to keep a list of what "image" changes happened when. I think it's probably worth pulling this out of the NEWS.md for connect, and can just update the README.md. If we decide to maintain a NEWS.md for the content images, we can always just use date `# 2021-11-17`, etc. since there is no such thing as a "version" 😄

connect/NEWS.md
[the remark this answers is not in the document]
sagerb: Changes made.

connect/NEWS.md
sagerb: changes made.

aronatkins (approved):

colearendt (approved):
LGTM!

colearendt (changes requested):
Actually - take that back. Let's look at the build failures

colearendt (approved):
Nvm. The "pro" images fail to build in their first PR because they depend on pushed "base" images 😢  That is an unfortunate case

PR #236, opened 2021-11-18
Add python dependencies for pyenv to each image

blairj09:
This PR introduces the dependencies outlined [here](https://github.com/pyenv/pyenv/wiki#suggested-build-environment) for using `pyenv` with linux.


blairj09:
https://github.com/rstudio/sol-eng-demo-server/pull/86

PR #237, opened 2021-11-18
Quick content Readme update

sagerb:
Split the generation of the base and pro YAML file as well as a quick note to indicate that `matrix.json` should be kept in sync with `build-images.sh`.


aronatkins (approved):

colearendt (approved):
LGTM!

PR #238, opened 2021-11-19
refactor build-content to build and push in the same job

colearendt:
artifact size is causing an issue with pro content images

### DIFF
--- a/.github/workflows/build-content.yaml
+++ b/.github/workflows/build-content.yaml
@@ -75,6 +75,28 @@ jobs:
         run: |
           docker image ls
 
+      - name: Login to Docker Hub
+        if: ${{ github.ref == 'refs/heads/main' }}
+        uses: docker/login-action@v1
+        with:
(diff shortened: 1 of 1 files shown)


dbkegley:
nit: we probably don't need the tar output from the builds anymore

https://github.com/rstudio/rstudio-docker-products/blob/5a1d5eeadd4e35f857f7e999ff5a5bcff7a24446/.github/workflows/build-content.yaml#L60

https://github.com/rstudio/rstudio-docker-products/blob/5a1d5eeadd4e35f857f7e999ff5a5bcff7a24446/.github/workflows/build-content.yaml#L116

dbkegley (approved):

colearendt:
I actually think those are used by the `digest` command:

https://github.com/rstudio/rstudio-docker-products/blob/5a1d5eeadd4e35f857f7e999ff5a5bcff7a24446/.github/workflows/build-content.yaml#L72

PR #239, opened 2021-11-20
add additional needed args

colearendt:
remove outputs


colearendt:
Finishes out #238 by loading the docker images into the session

PR #245, opened 2021-12-09
add libgit2 sysdep

colearendt:
this is important to the `gert` package, which is increasingly used in
the `tidyverse`


dbkegley (approved):

aronatkins (approved):
@sagerb - do we need to suggest this dependency elsewhere, as well?

colearendt:
I need to do some more digging on this... the `libgit2` dependency on `libcurl-gnu-dev` conflicts with other dependencies on `libcurl-openssl-dev`, so moving to a draft until we decide the best way forwards.

colearendt:
Turns out we don't need this! (I don't think)

PR #246, opened 2021-12-11
delete pro-drivers rpm and deb after installation 

michaelmayer2:
in order to keep docker container size under control.

It seems we are leaving the pro-drivers install packages lying around in the container - since I am not aware this would be any useful I am suggesting to remove them.


michaelmayer2:
Opening a clean one in a moment.

michaelmayer2:
Sorry for the mess - opening a clean pull request in a moment

PR #247, opened 2021-12-12
delete rstudio-drivers deb and rpm from r-session-complete after install

michaelmayer2:
We are installing our pro-drivers in r-session-complete but do not seem to remove the deb and rpm package once installed. Will save a few hundred of MB of storage space in the docker container.


colearendt:
A bit overkill, since you work here, but would you mind filling out the CLA just to make the CI happy? 😄

michaelmayer2:
Absolutely - I already had signed it but not added the respective eMail address to my github account... Done now and after rechecking it now all it green and good to go.

PR #258, opened 2022-01-18
Add krb5-user to rstudio workbench and r-session images

alessap:
Adding krb5-user to the ubuntu based images allows for a Active Directory user to authenticate using kerberos tickets within either the RStudio Workbench or the R session image.


colearendt (commented):
Thanks for submitting this!! I have to admit that I don't quite understand the version bump 😅  

The version number there is the RStudio Workbench version number, and we typically pin to the latest released version of RStudio Workbench. I suspect it is best to revert that change, unless I am missing something 😄

alessap:
@colearendt I had misunderstood what that version number represented. I thought it was the docker image version and as it was updated I though it required a version bump. I have reverted the commit as you suggested.

colearendt (approved):
LGTM! Just a note that we may need to revisit this when we move off of ubuntu bionic, because I recall hearing that `krb5` is deprecated 😑 IIRC it is deprecated in favor of `sssd`, which is much less preferable since it is a running service. 🤷‍♂️ Oh well. A problem for future us 😄

PR #263, opened 2022-01-31
change azure ml image version

colearendt:
- strip the build metadata off of SAFE_VERSION
- create a reference variable for `SHA_SHORT`
- start building `SHA_SHORT` tags for all images

For example:
- `rstudio-workbench` image w/ RSW version `2021.09.2-382.pro1`:
    - `latest`
    - `2021.09.2-382.pro1`
    - `2021.09.2-382.pro1--abcd123`
- `rstudio-connect` image w/ RSC version `2021.12.1`:
    - `latest`
    - `2021.12.1`
    - `2021.12.1--abcd123`
- `rstudio-workbench-for-microsoft-azureml` image w/ RSW version `2021.09.2-382.pro1`:
    - `latest`
    - `2021.09.2`
    - `2021.09.2--abcd123`

### DIFF
--- a/.github/workflows/build-latest.yaml
+++ b/.github/workflows/build-latest.yaml
@@ -67,10 +67,10 @@ jobs:
           tags: |
             rstudio/${{ matrix.config.product }}:${{ matrix.config.tag_prefix }}${{ matrix.config.version }}
             rstudio/${{ matrix.config.product }}:${{ matrix.config.tag_prefix }}${{ steps.version.outputs.SAFE_VERSION }}
-            rstudio/${{ matrix.config.product }}:${{ matrix.config.tag_prefix }}${{ steps.version.outputs.SAFE_VERSION }}{{ steps.version.outputs.SHA_SHORT }}
+            rstudio/${{ matrix.config.product }}:${{ matrix.config.tag_prefix }}${{ steps.version.outputs.SAFE_VERSION }}${{ steps.version.outputs.SHA_SHORT }}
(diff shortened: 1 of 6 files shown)


.github/workflows/build-latest.yaml
```diff
@@ -38,6 +38,14 @@ jobs:
           # NOTE: --local ensures that "latest" builds use the version in our Makefile (instead of latest from the web)
           VERSION=`./get-version.py ${{ matrix.config.product }} --type=${{ matrix.config.version }} --local`
           SAFE_VERSION=`echo -n "$VERSION" | sed 's/+/-/g'`
+
+          # strip build metadata for azure builds
+          if [[ "true" == "${{ matrix.config.product == 'rstudio-workbench-for-microsoft-azure-ml' }}" ]]; then
```
aronatkins, line 43: Is this double-comparison necessary? It certainly made me look twice!

Something like the following is how I generally check for value equivalence:

```bash
if [ "${val}" = "expected" ] ; then echo matching ; fi
```

Allowing for GHA variable expansion, that would probably be:

```bash
if [ "${{ val }}" = "expected" ] ; then echo matching ; fi
```

Is this sort of approach not possible here?

aronatkins (approved):
Could you give examples in the PR description showing the complete tags for different permutations?

.github/workflows/build-latest.yaml
colearendt: Ahh good point! I probably can just compare the `matrix.config.product` output. I think I stole this from somewhere it wasn't possible / there was more logic on the GHA side / expression. I do like simpler though 😄

PR #267, opened 2022-02-12
RSC: write RSC logs to file by default

pat-s:
When the license deactivation trap is triggered, usually something went wrong during startup.
In some scenarios, for example when running on k8s, inspecting logs is not so easy as the pod is not accessible after a failed startup.

This proposed change 

1. Writes the logs to a file and allows users to make them persistent across container updates (for simple docker setups).
1. Allows to see what might have caused the startup failure by printing the tail of the logs before executing the license deactivation trap

(NB: This is our setup we use for quite some time. I have not yet found an advantage sending logs to stdout - but maybe there is? 🤔 )

Please see this PR as a proposal to be discussed and not as something that blocks us :)

### DIFF
--- a/connect/startup.sh
+++ b/connect/startup.sh
@@ -5,6 +5,7 @@ set -x
 
 # Deactivate license when it exists
 deactivate() {
+    tail -n 200 /var/log/rstudio-connect.log
     echo "Deactivating license ..."
     /opt/rstudio-connect/bin/license-manager deactivate >/dev/null 2>&1
 }
@@ -25,4 +26,4 @@ unset RSC_LICENSE
 unset RSC_LICENSE_SERVER
 
 # Start RStudio Connect
-/opt/rstudio-connect/bin/connect --config /etc/rstudio-connect/rstudio-connect.gcfg
+/opt/rstudio-connect/bin/connect --config /etc/rstudio-connect/rstudio-connect.gcfg >>/var/log/rstudio-connect.log 2>&1


colearendt:
Hey @pat-s ! Thanks for submitting this / starting the discussion!

While there will definitely be those (like yourself) who prefer this approach and write logs to a file-share or other persistent location, there are a few reasons we have opted for stdout/stderr:
- most tooling inside of docker container orchestrators (Kubernetes, ECS, etc.) depend on logs coming to stdout/stderr
- the filesystem in many cases is considered "ephemeral" and will be destroyed when the container dies (in some cases, writing to the ephemeral disk at all is prohibited)
- similarly, if too much data is written to the ephemeral filesystem, orchestrators like Kubernetes will evict the pods for utilizing too much disk space
- as a result, there are mechanisms in place (i.e. in the docker daemon) to persist logs after containers die. For instance, in Kubernetes you can use `kubectl logs -p` to look at previous container runs, and using log aggregators (open source and paid) like DataDog, LogStash, etc. are quite common. These tools will often run on the Kubernetes / host nodes themselves and forward the logs from the docker daemon / all of its containers

As a result, we are unlikely to set up the defaults for our containers to log to the filesystem in the foreseeable future. Our goal is to integrate into the "paradigm" of containers as seamlessly as possible, and thereby as many third-party tools as possible. Our hope is to provide better guidance and patterns on such best practices in the future.

However, what you have here is a perfectly legitimate pattern to utilize in customer environments. Further, we will continue to look for ways to simplify the process of maintaining / debugging our products, so that it is easier to understand when things go wrong / why they went wrong / what to do about it.

I hope that helps 😄  Definitely happy to keep this conversation going!

colearendt:
Closing for now! Please let us know if you have other thoughts!

PR #268, opened 2022-02-15
Add tooling to only build diffs

colearendt:
- Add `get-diffs.py` that filters a `matrix.json` on the `dirs` parameter that have `diffs` (also allows specifying `--dirs` or returning `--all`)
- Convert `build-latest` to use `matrix.json`

TODO:
- Filter `matrix.json` and `matrix-preview.json` based on diffs (in the GHA)
- create / improve docstrings / etc.
- Decide whether to "_only test diffs & only build diffs_" or "_test all and only build diffs_"

### DIFF
--- a/get-diffs.py
+++ b/get-diffs.py
@@ -23,9 +23,7 @@ import subprocess
 def uniq_list(input: typing.List[typing.Any]):
     output = []
     for item in input:
-        if item in output:
-            continue
-        else:
+        if item not in output:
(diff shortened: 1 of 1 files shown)


atheriel:
My recommendation: since there are no interdependencies and you don't want to gate PRs on unrelated failures, I think testing only diffs is appropriate.

atheriel (approved):

PR #269, opened 2022-02-16
Make it possible to run the RSPM image as a non-root user

atheriel:
The official Ubuntu packages for RSPM set up an `rstudio-pm` user to run the service. However, the existing Docker image could not be run as this user for two reasons:

* The existing Dockerfile creates some directories required at runtime as root; and

* The existing installation step does not use [userspace license activation](https://docs.rstudio.com/rspm/admin/appendix/minimal-root-use-installation/#licensing), so later attempts to activate a license as a non-root user will fail.

This commit fixes both these issues. Images still default to running as root (for now, to preserve compatibility), but they can also be run as `rstudio-pm` now as well.


atheriel:
Build is failing but works for me locally. Will investigate.

colearendt:
Ugh. License initialization failing b/c a trial is not available. I'm not sure whether RSPM does an initialization automatically as a part of activation (in which case we can skip it)? IIRC RSW does that. If not, we may need to figure out how to initialize the license manager at runtime (entrypoint script?). 

It's also possible that even a failed initialization is still "good enough" to use the image 🤷‍♂️

atheriel:
> I'm not sure whether RSPM does an initialization automatically as a part of activation

Yeah, that's what I'm skipping. The CI logs also seem to indicate that it's being skipped correctly. I'm not using `docker buildx` locally, that might be related.

bdeitte:
Exciting to see!  Could you add a note on how to do this to the docs within https://github.com/rstudio/rstudio-docker-products#privileged-containers?

colearendt (commented):
This looks great! Maybe target the `dev` branch so we can build a few test containers to ensure backwards compatibility before we send this out to the world?

Also, might be worth adding some `goss` tests to protect this functionality from regressions if you are so inclined 😄

package-manager/Dockerfile
```diff
@@ -47,7 +46,21 @@ RUN apt-get update --fix-missing \
 ENV RSPM_LICENSE ""
 ENV RSPM_LICENSE_SERVER ""
 COPY rstudio-pm.gcfg /etc/rstudio-pm/rstudio-pm.gcfg
-VOLUME ["/data"]
+
+# These directories must be owned by rstudio-pm:rstudio-pm for a container to
+# work correctly when run as this user. (The container can also be run as root,
+# which will have access to these directories anyway.)
+RUN mkdir -p /var/run/rstudio-pm /data/launcher_internal \
```
colearendt, line 53: Is this `/data/launcher_internal` directory cluster-wide state? 😮  It seems a little scary to chown it recursively if it is.

glin (commented):
Great change, I tried it out and it works well. Related to Brian's question, should we document how exactly you run the RSPM image as `rstudio-pm`? Is that just adding `-u rstudio-pm` to your `docker run` command?

There was only one minor issue I noticed - that mounting a data volume gets trickier when we run RSPM or the Launcher as `rstudio-pm`. If you run the image as root and a fresh data directory, the data dir gets created as root and the `rstudio-pm` Launcher user can't access it:
```sh
$ docker run -it --rm --privileged \
  -p 4242:4242 \
  -e RSPM_LICENSE=$RSPM_LICENSE \
  -v $PWD/data/rspm:/data \
  rstudio/rstudio-package-manager:2021.12.0-3

...
17 Feb 2022 18:27:16 [rstudio-launcher] INFO Bootstrapping plugin Local
17 Feb 2022 18:27:16 [rstudio-launcher] INFO Received stderr from plugin Local: Unable to open config file: /data/launcher_internal/0ff094c1b802/launcher.local.conf
17 Feb 2022 18:27:16 [rstudio-launcher] INFO Received stderr from plugin Local: 17 Feb 2022 18:27:16 [rstudio-local-launcher-0ff094c1b802] ERROR system error 2 (No such file or directory) [path: /data/launcher_internal/0ff094c1b802/launcher.local.conf, description: Unable to open config file: /data/launcher_internal/0ff094c1b802/launcher.local.conf]; OCCURRED AT rstudio::core::Error rstudio::core::FilePath::openForRead(std::shared_ptr<std::basic_istream<char> >&) const src/cpp/shared_core/FilePath.cpp:1426; LOGGED FROM: bool rstudio::core::program_options::parseConfigFile(rstudio_boost::program_options::variables_map&, const string&, const rstudio::core::program_options::OptionsDescription&, bool) src/cpp/core/ProgramOptions.cpp:136
17 Feb 2022 18:27:16 [rstudio-launcher] INFO Plugin Local exited with code 1
17 Feb 2022 18:27:16 [rstudio-launcher] ERROR Could not start plugin Local; LOGGED FROM: rstudio::core::Error rstudio::job_launcher::plugins::PluginManager::start() src/cpp/job_launcher/plugins/PluginManager.cpp:148
17 Feb 2022 18:27:16 [rstudio-launcher] ERROR system error 71 (Protocol error) [description: Failed to properly bootstrap plugin Local]; OCCURRED AT rstudio::core::Error rstudio::job_launcher::plugins::Plugin::bootstrap() src/cpp/job_launcher/plugins/Plugin.cpp:189; LOGGED FROM: int main(int, char* const*) src/cpp/job_launcher/LauncherMain.cpp:259
2022/02/17 18:27:16 Launcher process exited: exit status 1
2022/02/17 18:27:16 Launcher process will restart in 1 seconds
```
I'm not really familiar with Docker volumes, but it seems like mounting the volume wipes out the `chown -R rstudio-pm:rstudio-pm /data` done in the image.

And when running the image as `rstudio-pm`, the data dir isn't accessible at all, and I only got it working by pre-creating a data directory with 777 permissions.
```sh
$ docker run -it --rm --privileged \
  -p 4242:4242 \
  -e RSPM_LICENSE=$RSPM_LICENSE \
  -u rstudio-pm \
  -v $PWD/data/rspm:/data \
  rstudio/rstudio-package-manager:2021.12.0-3

...
2022/02/17 18:29:58 Using data directory: /data
2022/02/17 18:29:58 Creating database directory: /data/db
2022/02/17 18:29:58 Error: Unable to initialize the configuration: mkdir /data/db: permission denied
```

colearendt:
Yeah, @glin that is my concern as well. I think you could also solve it by `chown`ing the data directory (in environments like NFS where customers would be running). That is probably the potential breaking change that we need to worry about for switching the user from root to rstudio-pm and document clearly for users. Hopefully not an issue in this case though.

atheriel:
@colearendt @glin The volume permissions issue is [a fundamental limitation of Docker volumes](https://stackoverflow.com/questions/33681396/how-do-i-write-to-a-volume-container-as-non-root-in-docker), so I don't think this is unique to us. Kubernetes does not use Docker volumes and handles a lot of these permissions for you correctly, so I don't think it will create issues there. It's more for local users running `docker run -v` directly.

> I'm not really familiar with Docker volumes, but it seems like mounting the volume wipes out the chown -R rstudio-pm:rstudio-pm /data done in the image.

Yes, it does. Those commands are only there to make it possible to run the container without persistent storage.

package-manager/Dockerfile
atheriel: This only takes effect if there is no volume mount (i.e. no persistent storage), so I don't think it will be a concern for cluster setups, since they require persistent/shared storage.

No, there's actually an annoyingly specific reason you need this: if you are running as root (the default), the Launcher user is still `rstudio-pm` by default. On startup, RSPM will create a `launcher_internal` subdirectory, but if RSPM is run as root this will only be readable by root... and so the launcher will fail to start, because it can't read that directory -- and the startup will fail.

atheriel:
I think the current plan is to merge this first and rebase #270 on top afterwards.

colearendt (approved):
LGTM!

PR #270, opened 2022-02-17
RSPM: Run RSPM as rstudio-pm instead of root

tylfin:


package-manager/Dockerfile
```diff
@@ -27,12 +27,6 @@ ARG TINI_VERSION=0.18.0
 RUN curl -L -o /usr/local/bin/tini https://github.com/krallin/tini/releases/download/v${TINI_VERSION}/tini && \
     chmod +x /usr/local/bin/tini
 
-COPY startup.sh /usr/local/bin/startup.sh
-RUN chmod +x /usr/local/bin/startup.sh
-
-EXPOSE 4242/tcp
```
tylfin, line 33: Unsetting these fields because it's impossible for users that pull `FROM ...` to undo them

atheriel:
Is this a re-go at #269?

tylfin:
I believe this supercedes https://github.com/rstudio/rstudio-docker-products/pull/269

package-manager/Dockerfile
atheriel: Agreed. (Also, these statements have no effect -- they are just metadata, users still need to map ports directly.)

package-manager/README.md
```diff
@@ -34,38 +38,33 @@ Open [http://localhost:4242](http://localhost:4242) to access RStudio Package Ma
...
-
-See a complete example of that file at `package-manager/rstudio-pm.gcfg`.
-
-#### Persistent Data
+## Persistent Data
 
 In order to persist Package Manager data between container restarts configure RSC `Server.DataDir` option to go to
 a persistent volume. The included configuration file expects a persistent volume from the host machine or your docker
-orchestration system to be available at `/data`. Should you wish to move this to a different path, you can change the
+orchestration system to be available at `/var/lib/rstudio-pm`. Should you wish to move this to a different path, you can change the
```
atheriel, line 64: I think this is a good change. `/data` is kind of random, and I know we ignore it in the Helm charts, so it's really only for local testing.

package-manager/rstudio-pm-float.gcfg
```diff
@@ -1,14 +1,56 @@
...
+; Address = https://rstudio-pm.company.com
+Address =
+;
+; Git sources require a configured R installation. R is often installed at `/usr/lib/R`
+; or `/usr/lib64/R`.
+RVersion = /opt/R/3.6.2/
+;
+; Customize the data directory if necessary. This is where all packages and metadata are
+; stored by default. Refer to Admin Guide for details.
+;DataDir = /mnt/rspm/data
```
atheriel, line 18: Should this be `/var/lib/rstudio-pm`?

glin:
Allowing the image to run as a non-root user alone feels very safe to me, as done in https://github.com/rstudio/rstudio-docker-products/pull/269. But this PR is different, running the image as both non-root and unprivileged, which means Git sandboxing must be disabled.

Are we okay with disabling Git sandboxing by default? We talked about the risks of that earlier at https://github.com/rstudio/package-manager/pull/6390#issuecomment-905579728. Maybe it's safer to run unprivileged than to disable sandboxing, but at least the `--privileged` flag is more explicit than having sandboxing disabled in the default config.

Jon had suggested changing `Launcher.ServerUser` to a different user from `rstudio-pm`, so at least the Launcher jobs would be isolated from RSPM data. That might be a simple enough improvement. Or we could figure out how to get sandboxing to work in unprivileged containers, as @atheriel mentions at https://github.com/rstudio/package-manager/issues/7443.

atheriel:
Tyler and I spoke about this a bit as well. Currently I think the need to use `--privileged` is very unsafe and I think non-sandboxed builds are a better security option for most users. However, I'm not sure disabling the sandbox is the right default, I'd rather have users make an explicit choice to use it (with `--privileged`, at least for now) or not (by setting the config entry, or potential an environment variable in future releases).

But in the defence of this current PR, the changes to the README make it very clear we're disabling sandboxing.

tylfin:
@glin This PR is meant to bring the Package Manager image back into the recommended configuration and making it possible to migrate the dogfood sites to.

> Jon had suggested changing Launcher.ServerUser to a different user from rstudio-pm, so at least the Launcher jobs would be isolated from RSPM data. That might be a simple enough improvement.

This seems like the right answer to me, is there anything special we need to do to facilitate that again?

glin:
@atheriel @tylfin That reasoning sounds good to me. I would prefer it to be an explicit choice, or at least explain the difference between `--privileged`/sandboxed vs. unprivileged/unsandboxed so users can choose. Right now, the README suggests that the default is unsandboxed and insecure, but doesn't explain that sandboxed could actually be more insecure because of the `--privileged` limitation.

@tylfin For `Launcher.ServerUser`, I'm not sure, I tried it a while back but couldn't get it to work because of the `launcher_internal` permissions issue (https://github.com/rstudio/package-manager/issues/6388#issuecomment-905801455). I don't think I tested whether a manual chown of the `launcher_internal` directory would work around it, and maybe it does. You can try it out here. I think we'll also need a separate `Launcher.AdminGroup` since that defaults to `rstudio-pm` as well.

glin:
I tried using a different `Launcher.ServerUser` and it unfortunately doesn't work :cry:. First, both RSPM and the Launcher need read/write access to the `launcher_internal` directory, which is annoying but can be worked around. The big issue is that RSPM requires `Launcher.ServerUser` to be the same as the RSPM user if RSPM isn't running as root:
```sh
2022/02/17 23:41:02 Error: Unable to initialize the configuration: The service is currently running as 'rstudio-pm' but Launcher.ServerUser is configured to run as rstudio-pm-launcher. If the service is not running as root, we cannot run the launcher as a different user.
```
https://github.com/rstudio/package-manager/blob/d130806e15b0114e450f84616f649774e81d3bd7/src/rspm/config/config.go#L1253-L1259

Running RSPM as `root` with Launcher as `rstudio-pm` might be better, but then you hit another known issue with `root` RSPM requiring `--privileged` or other Docker security flags (https://github.com/rstudio/package-manager/pull/6390#issuecomment-1042323544).

package-manager/rstudio-pm-float.gcfg
tylfin: I think this is meant to serve as an example of changing the DataDir to a specific mount, this is brought in from the file we ship with RSPM https://github.com/rstudio/package-manager/blob/main/config/rstudio-pm.gcfg#L18

tylfin:
> But for the specific goal of serving our dogfood sites, isn't the sandboxing change to the default config irrelevant because those sites will use non-default configurations anyway?

Oh that's true, I also wanted things to just work out of the box for customers, but if there's a better way to accomplish it I'm game

tylfin:
**Note:** Do not merge. These breaking changes are scheduled for the next official RSPM release.

package-manager/Dockerfile
```diff
@@ -26,11 +25,7 @@ ARG TINI_VERSION=0.18.0
...
-EXPOSE 2112/tcp
+ENV RSTUDIO_INSTALL_NO_LICENSE_INITIALIZATION 1
```
atheriel, line 28: I think this is already set in the `RUN apt-get...` statement below. We might want to remove this `ENV` statement so that this installation-specific variable does not get leaked into running containers.

package-manager/Dockerfile
```diff
@@ -43,24 +38,22 @@ RUN apt-get update --fix-missing \
     && rm -rf /var/lib/apt/lists/* \
     && ln -s /opt/rstudio-pm/bin/rspm /usr/local/bin/rspm
 
-ENV RSPM_LICENSE ""
-ENV RSPM_LICENSE_SERVER ""
-COPY rstudio-pm.gcfg /etc/rstudio-pm/rstudio-pm.gcfg
+# Add run script and set permissions -------------------------------------------#
+COPY startup.sh /usr/local/bin/startup.sh
+RUN mkdir -p /var/run/rstudio-pm && \
+    chmod +x /usr/local/bin/startup.sh && \
```
atheriel, line 44: Nit: For consistency, most of the other lines in this file have `&&` as a line prefix instead of a suffix.

package-manager/README.md
```diff
@@ -76,34 +76,34 @@ Using the RStudio Package Manager docker image requires to have a valid License.
...
     -p 4242:4242 \
     -v $PWD/package-manager/rstudio-pm.gcfg:/etc/rstudio-pm/rstudio-pm.gcfg \
     -e RSPM_LICENSE=$RSPM_LICENSE \
     rstudio/rstudio-package-manager:latest
 
 # Run with persistent data and using an external configuration
-docker run -it --privileged \
+docker run -it \
     -p 4242:4242 \
     -v $PWD/data/rspm:/data \
```
atheriel, line 108: I think `/data` needs to become `/var/lib/studio-pm/` here.

atheriel, line 108: I think `/data` needs to become `/var/lib/studio-pm/` here.

package-manager/rstudio-pm.gcfg
```diff
@@ -1,12 +1,53 @@
...
+;DataDir = /mnt/rspm/data
 
-[Git]
-BuilderDir = "/tmp/git"
+[HTTP]
+; RStudio Package Manager will listen on this network address for HTTP connections.
+Listen = :4242
 
-[Metrics]
-Enabled = true
```
atheriel, line 12: We're turning off metrics here, which might be a breaking change for users.

tylfin:
cc @atheriel now that 2022.04 is out, can we revisit this? Is there anything blocking its release?

package-manager/Dockerfile
```diff
@@ -37,24 +31,22 @@ RUN apt-get update --fix-missing \
     && rm -rf /var/lib/apt/lists/* \
     && ln -s /opt/rstudio-pm/bin/rspm /usr/local/bin/rspm
 
-ENV RSPM_LICENSE ""
-ENV RSPM_LICENSE_SERVER ""
-COPY rstudio-pm.gcfg /etc/rstudio-pm/rstudio-pm.gcfg
+# Add run script and set permissions -------------------------------------------#
+COPY startup.sh /usr/local/bin/startup.sh
```
atheriel, line 35: This got lost in the rebase. It should be

```
COPY --chmod=0775 startup.sh /usr/local/bin/startup.sh
```

with the `chmod +x /usr/local/bin/startup.sh && chown rstudio-pm:rstudio-pm /usr/local/bin/startup.sh` lines below removed.

package-manager/Dockerfile
```diff
@@ -37,24 +31,22 @@ RUN apt-get update --fix-missing \
...
+RUN mkdir -p /var/run/rstudio-pm \
+    && chmod +x /usr/local/bin/startup.sh \
+    && chown rstudio-pm:rstudio-pm /usr/local/bin/startup.sh \
+    && chown -R rstudio-pm:rstudio-pm /var/run/rstudio-pm \
+    chsh -s /bin/bash rstudio-pm
```
atheriel, line 40: Is the purpose of this to make it possible to `su - rstudio-pm` when logged in as `root`? Do we need that? Otherwise I'm not sure of the purpose, because the user's shell setting does not affect one's ability to `kubectl exec ... bask` or `docker exec ... bash` (as far as I know).

package-manager/Dockerfile
```diff
@@ -37,24 +31,22 @@ RUN apt-get update --fix-missing \
...
 
-ENV RSPM_LICENSE ""
-ENV RSPM_LICENSE_SERVER ""
-COPY rstudio-pm.gcfg /etc/rstudio-pm/rstudio-pm.gcfg
+# Add run script and set permissions -------------------------------------------#
+COPY startup.sh /usr/local/bin/startup.sh
+RUN mkdir -p /var/run/rstudio-pm \
+    && chmod +x /usr/local/bin/startup.sh \
+    && chown rstudio-pm:rstudio-pm /usr/local/bin/startup.sh \
+    && chown -R rstudio-pm:rstudio-pm /var/run/rstudio-pm \
```
atheriel, line 39: Nit: I think the post-install changes to create the `/var/run/rstudio-pm` directory are live in the latest release, so these lines may not longer be required at all. We can change them now or later.

atheriel (approved):

package-manager/README.md
```diff
@@ -14,8 +14,13 @@ Docker images for RStudio Professional Products
    changes. We
    provide [instructions for building](https://github.com/rstudio/rstudio-docker-products#instructions-for-building) for
    these cases.
+1. The Package Manager image uses the `No Sandbox` option documented
+   [here](https://docs.rstudio.com/rspm/admin/process-management/#process-management-sandboxing) by default, if you need
+   a more secure option for configuring
+   [Git-related package builds](https://docs.rstudio.com/rspm/admin/building-packages/) we recommend using a system with
+   sandboxing enabled.
```
glin, line 21: Could we add a direct link to the instructions for running RSPM in an unprivileged container with sandboxing enabled? That should mostly be the "User Namespace Support - Docker" section - https://docs.rstudio.com/rspm/admin/process-management/#docker. The Git building docs are still kind of hard to navigate and in need of some reorganization, so it would be nice to save users some trouble with finding that.

glin (approved):

atheriel:
@tylfin Nothing I am aware of at this point. LGTM.

colearendt:
This definitely strikes me as something that we should throw into `NEWS.md` 😄

package-manager/Dockerfile
```diff
@@ -1,6 +1,5 @@
 ARG R_VERSION=3.6.2
 FROM rstudio/r-base:${R_VERSION}-bionic
-LABEL maintainer="RStudio Docker <docker@rstudio.com>"
```
colearendt, line 3: Why did we remove the maintainer label? Did I miss that? This is a google group IIRC - we can always forward things as needed to the appropriate teams. But I do think that having some maintainer is important

package-manager/README.md
```diff
@@ -34,38 +39,33 @@ Open [http://localhost:4242](http://localhost:4242) to access RStudio Package Ma
...
-Note that running the RStudio Package Manager Docker image requires the container to run using the `--privileged` flag
-and a valid RStudio Package Manager license.
+Note that running the RStudio Package Manager Docker image requires a valid RStudio Package Manager license.
 
 This container includes:
 
-1. R 3.6.1
-1. RStudio Package Manager
+1. R 3.6.2
+2. RStudio Package Manager
```
colearendt, line 49: Just a note that markdown (maybe just GitHub markdown?) by convention will automatically increment, even if you don't increment your numbered list yourself (i.e. 1., 1., 1. is valid and will display as 1., 2., 3.)

package-manager/rstudio-pm.gcfg
colearendt, line 12: Is there a reason we are disabling Metrics by default? Are y'all just switching to using the default config file?

colearendt (commented):
Just a few nits, in addition to my request for a `NEWS.md` entry 😄

colearendt (changes requested):
One more thought that I want to throw out there - how thoroughly have we tested this change? In particular, backwards compatibility with previous installations (and their Data Directories) and with existing installations on the helm chart?

Because we are not bumping the version of the product, anyone pegged to the 2022.04 version will unexpectedly get a new image and could encounter a surprising breakage. We can still do that if it's desirable, but IMO we should have a set of resolution steps handy to save ourselves a firefight 😄 Moreover, we could stage compatible changes with the helm chart, or coordinate this change with a version bump of RSPM (if there are any imminent) to make it smoother too.

Finally, we do have a `dev` branch that gets automatic CI / container builds. If we want to slow this merge process a smidge, we could always merge there first to get containers building and see how they behave in real environments (i.e. Colorado Staging, etc.). One note - RSW has some changes staged in dev presently, so merging directly from dev to main is probably not the best bet right now 😓 But we could always merge this same commit into main once we are happy with it.

Happy to discuss tomorrow - just wanted to be sure these angles were discussed before we merge!

package-manager/rstudio-pm.gcfg
tylfin: Yes, the goal of this PR and moving forward is to have the RSPM Docker image match the default RSPM installation as much as possible

PR #283, opened 2022-04-01
demonstrate how to cp source tarballs to add in rspm

dpastoor:
The result of following the instructions in the `package-manager/README.me` will result in a new demopkgs repo created:

Pasted image 20220328084739

### DIFF
--- a/package-manager/README.md
+++ b/package-manager/README.md
@@ -242,6 +242,70 @@ docker exec -it {container-id} /bin/bash
 Then please refer to the [Package Manager documentation](https://docs.posit.co/rspm/admin/) on how
 to [create and manage](https://docs.posit.co/rspm/admin/getting-started/configuration.html) your repositories.
 
+#### Example for adding a Workbench-built tarball
+
+For this example, a package, `demo1` was created with version 0.1.0 and a
+tarball was created using the build pane in pwb. This will write out the
+tarball on the host system at `data/pwb/rstudio/demo1_0.1.0.tar.gz`
+
+Copy the tarball into the `/tmp` dir using `docker cp`:
+
+Find the container id for the running package-manager container via `docker ps`:
+
+```bash
+❯ docker ps
+CONTAINER ID   IMAGE ...
+b8ae944b7f2d   rstudio/rstudio-package-manager:2021.12.0-3 ...
+```
+
+```bash
+# replace b8ae944b7f2d with your container ID
+docker cp data/pwb/rstudio/demo1_0.1.0.tar.gz b8ae944b7f2d:/tmp
+```
+
+now you can create the repo in rspm
+
+```bash
+root@1b6f4f5fc669:/# rspm create source --name=demopkgs
+Source 'demopkgs':
+  Type:  Local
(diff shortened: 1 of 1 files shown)


colearendt (commented):
Interesting! Thanks for writing this up! I'll have to take a more careful read through the docs.

One question: what are your thoughts around `docker cp` and whether that would more neatly solve the problem of having to add another volume and the overhead, confusion, and permissions issues that volumes can create when written from the host and read from the container? (IIRC volumes can be a notorious troublesome spot in linux, as well as in certain versions of docker desktop for mac)

dpastoor:
> what are your thoughts around `docker cp` 

my thoughts are - oh nice! I've never used that 😎  

Lemme give that a whirl. The case that I can think of that I want to make sure doesn't get wonky is we'd need the guidance to prob match up to copy into /data so the tarball will still ultimately land in the persistent dir, else if you docker-compose down then the db/everything else expecting the tarball would still live on and then might go sideways on start up next time. 

I concur that keeping things as simple as possible especially around what crosses the host boundaries is ideal.

colearendt:
Nice! It's worth noting that RSPM also "caches" the tarball, so once you have imported it, you can actually dispense with it 😄 We don't make that very clear in our docs, as far as I can tell, but if you are persisting the database and the data directory then there is really no need to keep track of the tarball - ephemeral storage is just fine 😄

dpastoor:
that is definitely an interesting tidbit to know!

dpastoor:
round 2 - trying flow with docker cp:

built demo1 pkg in the rsw container:

image

we see demo1_0.1.0.tar.gz kicked out in data/rsw/rstudio

```
data/rsw/rstudio🔒 on  rspm_mnt [!]
❯ ls
R  demo1  demo1_0.1.0.tar.gz
```

find container id:

```
❯ docker ps
CONTAINER ID   IMAGE      
1b6f4f5fc669   rstudio/rstudio-package-manager:2021.12.0-3  ...
```

copy from host into rspm image:

```
docker cp data/rsw/rstudio/demo1_0.1.0.tar.gz 1b6f4f5fc669:/tmp
```

cool deal its there:

```
❯ docker exec -it 1b6f4 /bin/bash
root@1b6f4f5fc669:/# ls /tmp
demo1_0.1.0.tar.gz
```

```
root@1b6f4f5fc669:/# rspm create source --name=demopkgs
Source 'demopkgs':
  Type:  Local
root@1b6f4f5fc669:/# rspm add --source=demopkgs --path='/tmp/demo1_0.1.0.tar.gz'
Added package '/tmp/demo1_0.1.0.tar.gz'.
root@1b6f4f5fc669:/# rspm create repo --name=demopkgs --description="demo package repo"
Repository: demopkgs - demo package repo - R
root@1b6f4f5fc669:/# rspm subscribe --repo=demopkgs --source=demopkgs
Repository: demopkgs
Sources:
--demopkgs (Local)
```

and looks like we're in business:

image

dpastoor:
@colearendt `docker cp` worked like a charm - renamed the PR and updated the docs to show the new flow. Demonstration of e2e steps + screenshots also in PR above

bschwedler (approved):

PR #285, opened 2022-04-04
Remove unnecessary apt dependencies for Package Manager

atheriel:
It seems that libssl and sudo are not actually required to install and run RSPM, so we can remove them. Further, gdebi-core is only used during installation, so we can install it temporarily during that step instead.

This has the following benefits:

* Fewer `RUN` steps, and fewer layers.

* Fewer system packages in the resulting image, making it 10M smaller.

* Improved build time, because we don't have two `apt`-related steps (and their attendant cache cleanup code).

### DIFF
--- a/package-manager/Dockerfile
+++ b/package-manager/Dockerfile
@@ -9,16 +9,6 @@ ENV LANGUAGE en_US:en
 ENV LC_ALL en_US.UTF-8
 ENV DEBIAN_FRONTEND=noninteractive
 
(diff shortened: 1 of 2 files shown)


glin (approved):

colearendt (approved):
Nice! LGTM!

Mind adding a quick NEWS entry, just for future reference? Can just include the date and the change for reference, especially if this affects any downstream use of the image

PR #287, opened 2022-04-05
Dockerfile changes for ClamAV Scanning

jonyoder:
This adds support for scanning the AzureML images with ClamAV. See https://github.com/rstudio/platform-team/issues/161.


atheriel:
Can we add a scan step to GitHub Actions instead of using (and then discarding) a multistage build?

jonyoder:
@atheriel I have no idea. I was surprised at how little information I found on virus scanning for docker images. This seemed like a relatively lightweight solution that
* doesn't involve figuring out how to scan inside an image from outside it, and
* doesn't clutter the image with antivirus software.

Did you have a particular solution in mind?

atheriel:
Also: @jonyoder don't we want to run this on the Azure image primarily? Or is that planned for a second PR?

jonyoder:
@atheriel is that the Dockerfile inside the helpers directory? I'm obviously not familiar with this repo yet.

atheriel:
@jonyoder Yep! (I realize you may have thought that Azure used Workbench as a base image -- it does not, they have their own custom thing.)

jonyoder:
Thanks! I'll focus on scanning for Azure only for now.

helper/workbench-for-microsoft-azure-ml/Dockerfile
```diff
@@ -178,3 +178,38 @@ EXPOSE 5559/tcp
...
+RUN mv /etc/clamav/freshclam.conf.new /etc/clamav/freshclam.conf
+
+# Update ClamAV
+RUN touch /clamupdate.log
+RUN chown clamav:clamav /clamupdate.log
+RUN freshclam --log=/clamupdate.log
+
+# Scan all but /sys for viruses. If this fails, the Docker build will
+# fail.
+RUN clamscan --recursive --infected --log=/clamscan.log --exclude-dir="^/sys" /
```
bschwedler, line 204: Similar comment here. Please combine down into as few layers as possible.

jonyoder:
@bschwedler I made the changes you suggested, and I also updated to use the new CDN URL for ClamAV updates. This needs a PR approval before we can merge it, so let me know if there's anything else I can do.

helper/workbench-for-microsoft-azure-ml/Dockerfile
```diff
@@ -178,3 +178,38 @@ EXPOSE 5559/tcp
...
+# Set up ClamAV
+RUN apt-get update && \
+    apt-get install -y \
+        clamav \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
+
+# Reconfigure freshclam to get database from our S3 location
+RUN cat /etc/clamav/freshclam.conf | grep -v "DatabaseMirror" > /etc/clamav/freshclam.conf.new && \
+    echo "DatabaseMirror ${CLAMAV_MIRROR}" >> /etc/clamav/freshclam.conf.new && \
```
colearendt, line 194: What is the effect of using this ARG before it is declared as an ARG? Shouldn't line 198 precede this set of commands? (Presuming that you are using variable interpretation because double quotes?)

helper/workbench-for-microsoft-azure-ml/Dockerfile
```diff
@@ -178,3 +178,38 @@ EXPOSE 5559/tcp
...
+
+
+# Ignore the intermediate 'clamav' multi-stage build step so we don't distribute
+# with ClamAV installed. Avoid adding changes in this stage unless they are related
+# to the ClamAV stage. Since the ClamAV stage may be cached, you can't guarantee
+# another full scan if you change anything beyond this point.
+FROM workbench AS final
+
+# Copy ClamAV scan logs so the end user can see them.
+COPY --from=clamav /clamscan.log /clamupdate.log /
```
colearendt, line 215: I must admit that it surprises me we wouldn't do this after the RUN command above, and finish with `FROM workbench as final`?

colearendt (commented):
Just a few nits - otherwise LGTM!

colearendt:
Also, I like the idea of figuring out how to "scan the image from outside it," if there is some way to mount components in. Having a generic way to scan with a GHA job would probably be useful. However, if that's something that doesn't already exist in the community, it makes sense that it could be a hefty lift to get sorted

helper/workbench-for-microsoft-azure-ml/Dockerfile
jonyoder: Oh, this was a total mistake. I'll fix it.

helper/workbench-for-microsoft-azure-ml/Dockerfile
jonyoder: The purpose here is to copy the logs to the final docker image. Think of it as a signature of sorts to prove that we actually scanned the image. If we moved this to above the final `FROM`, then the docker image wouldn't retain any record of ClamAV scanning.

jonyoder:
@colearendt I fixed the `ARG` position issue. Regarding logs, the `COPY FROM` actually copies the logs to the docker image, so we do have a record of the scan. Because of this, I don't think we need GHA artifacts.

colearendt (approved):
LGTM!

PR #289, opened 2022-04-06
Avoid the COPY + chmod pattern

atheriel:
Unfortunately, using `COPY` to move a file into an image followed by `RUN chmod` actually duplicates the file and creates an unnecessary image layer.

This commit makes use of the `--chmod` flag that was added to `COPY` for this exact reason in a few places. This flag is only supported by BuildKit, but that's what we use to build images anyway, so it seems like a reasonable win.


bschwedler (approved):

colearendt (commented):
Nice, thanks! TIL!

This may be silly / overkill, but I think a NEWS entry may be nice to have for each of these, in case someone is using / forking / extending these dockerfiles with a different builder 😬

colearendt (approved):
Looks awesome!!

PR #290, opened 2022-04-06
Install Package Manager with apt instead of gdebi

atheriel:
This has a few advantages:

* We don't have to install (and later remove) `gdebi`.

* `ap`t allows us to skip installing Recommended packages. RSPM depends on `git` and `git` recommends `less` and `ssh-client` (yes, really), so we were previously installing an SSH client into the image. With this change neither of these packages nor their transitive dependencies are installed.

This shaves about 8M off the resulting image, too.

### DIFF
--- a/package-manager/Dockerfile
+++ b/package-manager/Dockerfile
@@ -26,13 +26,13 @@ EXPOSE 2112/tcp
 ARG RSPM_VERSION=2021.12.0-3
 ARG RSPM_DOWNLOAD_URL=https://cdn.rstudio.com/package-manager/ubuntu/amd64
 RUN apt-get update --fix-missing \
-    && apt-get install -y --no-install-recommends gdebi-core dpkg-sig \
+    && apt-get install -y --no-install-recommends dpkg-sig \
     && curl -O ${RSPM_DOWNLOAD_URL}/rstudio-pm_${RSPM_VERSION}_amd64.deb \
     && gpg --keyserver keyserver.ubuntu.com --recv-keys 3F32EE77E331692F \
     && dpkg-sig --verify rstudio-pm_${RSPM_VERSION}_amd64.deb \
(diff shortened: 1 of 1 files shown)


glin (commented):
The license manager doesn't seem to work with this change, as the Ubuntu 18 build of it requires `libssl1.0.0`?
```sh
/opt/rstudio-pm/bin/../ext/activation/license-manager: error while loading shared libraries: libcrypto.so.1.0.0: cannot open shared object file: No such file or directory
```

I didn't realize/forgot that RSPM actually does require libssl via license-manager and other RSW dependencies - rsandbox, launcher. `libssl1.0.0` seems to be coming in as a Recommends of git, so that's not great, we should make that an explicit dependency.

And also, the next release of RSPM on Ubuntu 18 uses RSW dependencies built against `libssl1.1`, so there's some maintenance pain with including libssl manually (e.g. https://github.com/rstudio/rstudio-docker-products/pull/285).

What do you think of keeping the gdebi change, but not using `--no-install-recommends` for now? Then we could fix the dependency issues with the RSPM packages for the next release.

atheriel:
Ah, so really there was an issue with #285 it was just masked by getting pulled in with `git`. I'd prefer explicitly installing `libssl1.0.0` for now, I think, and then fix it once the upstream dependencies are corrected. Reverting back to `gdebi` with a reminder to remove it in the future seems to muddy things even further.

shepherdjerred:
This would be nice to do. Using `gdebi` in our official install guide produces a warning that may mislead customers into thinking something is wrong.

https://github.com/rstudio/package-manager/issues/11698

bschwedler:
We should be able to do this now that Ubuntu 18.04 is EOL.

bschwedler:
Closing this out since we are using now `apt` for all Debian package installations:
https://github.com/rstudio/rstudio-docker-products/blob/4fd9cb36d8ff1580347f51aec2bee5a6b5c6bb4e/package-manager/Dockerfile.ubuntu2204#L29

PR #294, opened 2022-04-19
Update for RStudio Package Manager 2022.04.0-7

shepherdjerred:
Update to RStudio Package Manager 2022.04.0-7


colearendt (changes requested):
The "source" of these versions is upstream in the `Makefile` (and we have a `make update-versions` target that should help with this / ensure the appropriate places are all accounted for)! (This is also the version used in CI)

shepherdjerred:
Thanks Cole! I updated the `Makefile` version and ran `make update-versions`.

colearendt (approved):
LGTM! Thanks! 🎉

PR #300, opened 2022-04-29
Remove the unused pandoc and TeX installations from the RSPM image

atheriel:
Previously the image was based on `rstudio/r-base`, which includes TinyTeX and pandoc. However RSPM's Git builds do not build package manuals or vignettes, so we don't actually need these programs.

This shaves about 300MB off the image size.

The upstream `rstudio/r-base` image also does the following, which we replicate directly instead:

* Install R from RStudio's own packages.

* Set `TZ=UTC`.

* Set the locale.

### DIFF
--- a/package-manager/Dockerfile
+++ b/package-manager/Dockerfile
@@ -1,13 +1,28 @@
-ARG R_VERSION=3.6.2
-FROM rstudio/r-base:${R_VERSION}-bionic
+FROM ubuntu:bionic
 LABEL maintainer="RStudio Docker <docker@rstudio.com>"
+
+# System dependendies and an R installation for package builds -----------------#
+ARG R_VERSION=3.6.2
+ARG R_DOWNLOAD_URL=https://cdn.rstudio.com/r/ubuntu-1804/pkgs
+RUN apt-get update -qq \
+    && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+    ca-certificates \
+    curl \
+    locales \
+    && curl -O ${R_DOWNLOAD_URL}/r-${R_VERSION}_1_amd64.deb \
+    && DEBIAN_FRONTEND=noninteractive apt-get install -f -y ./r-${R_VERSION}_1_amd64.deb \
+    && rm r-${R_VERSION}_1_amd64.deb \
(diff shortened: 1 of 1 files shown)


package-manager/Dockerfile
```diff
@@ -1,13 +1,28 @@
...
+    ca-certificates \
+    curl \
+    locales \
+    && curl -O ${R_DOWNLOAD_URL}/r-${R_VERSION}_1_amd64.deb \
+    && DEBIAN_FRONTEND=noninteractive apt-get install -f -y ./r-${R_VERSION}_1_amd64.deb \
+    && rm r-${R_VERSION}_1_amd64.deb \
+    && rm -rf /var/lib/apt/lists/* \
+    && ln -s /opt/R/${R_VERSION}/bin/R /usr/bin/R \
+    && ln -s /opt/R/${R_VERSION}/bin/Rscript /usr/bin/Rscript \
+    && ln -s /opt/R/${R_VERSION}/lib/R /usr/lib/R
```
glin, line 18: Looking at this now, I think it was a mistake that we symlinked R to the OS default `/usr/bin` instead of `/usr/local/bin` in the rstudio/r-base images. The symlink to `/usr/lib` also isn't necessary at all, I copy pasted that from somewhere else :grimacing:. I don't know if any users would ever want to switch their R versions here, but using `/usr/local/bin` would be nice for consistency with the instructions at https://docs.rstudio.com/resources/install-r/#create-a-symlink-to-r

glin: Or we could remove it since RSPM doesn't need R to be on the PATH. Unless users might be running R scripts in this image.

glin (approved):
Looks good, this is a great idea. Should we add a NEWS item as well?

colearendt (commented):
I think this is a great call, along with @glin 's recommendations. In particular, I think a NEWS entry would definitely be worthwhile

package-manager/Dockerfile
atheriel: Definitely not the happy path. I'll remove these symlinks.

atheriel:
I'm going to wait for #270 to be merged before creating more churn with this one.

PR #309, opened 2022-05-06
EICAR test

jonyoder:
This is to prove that the Azure ML image is scanned for viruses and fails when one is found. Don't merge this; I'll delete it after we prove the build fails.


jonyoder:
Closing since the test was successful

PR #310, opened 2022-05-09
feat(connect): add active directory prerequisites

starlightromero:
Add `libnss-sss` and `libpam-sss` packages added so Connect can run tasks as Active Directory users


starlightromero:
@colearendt @glin

colearendt:
Thanks for the PR @starlightromero ! Apologies for the delay getting to this - can you say a bit more about why these libraries in particular are needed / sufficient? Are you running the `sssd` process as a sidecar? How do the `pam` and `nss` modules suffice to provision users inside of the container?

A bit more background would be helpful for us understanding your use case / architecture a bit more!

starlightromero:
I appreciate you getting back to me. We are running SSSD as a sidecar and have a volume mount that is for the entire `var/lib/sss` directory

jgoldschrafe:
Small addendum for clarity: we're running Okta as our identity source for user login through the UI, but we also need consistent UID/GID mapping for user IDs running Connect jobs. We have SSSD wired to provide this from Active Directory via the (K8s) sidecar that @starlightromero mentioned, and the sidecar works great, but getting the requisite (and very general-purpose!) PAM/nsswitch libraries into the container currently requires us to maintain a custom image for Connect.

colearendt:
That makes a lot of sense! Thanks for the background here, y'all! I had never considered that you could use sssd as a sidecar with the appropriate libraries and volume mount for the socket - that's brilliant! 

In fact, this may be something we take a look at for Workbench in the future (where we currently run `sssd` inside of the container out of current necessity). I can't see any reason to oppose these libraries on those grounds! 

I just kicked off the build, and then we should be able to get this merged! This could be a very useful pattern to share w/ our customers - are there any keys that y'all have found being successful in this approach? Is this a pattern you use for other services as well, or was Connect the first?

starlightromero:
I believe this is the first time we have used this approach. We follow this same pattern with Workbench. 

As far as keys to success, @jgoldschrafe may be better suited to answer.

starlightromero:
@colearendt Looks like it's wanting another approval

colearendt (approved):
Woops - sorry for the delay here! Looking forward to having this merged!

PR #315, opened 2022-05-19
Build less frequently

### LINKED ISSUE #255: build images less, and possibly with additional versioning
We recently observed the content-base images get rebuilt when the source for those images had not changed. That pulled in some new Python package versions, which were incompatible (for reasons). See https://github.com/rstudio/rstudio-docker-products/pull/254 for details.

Some potential improvements:
* Images rebuild only when explicitly triggered or, perhaps, when the source associated with those images changes.
* Images receive a versioned label in addition to their current, floating label.

colearendt:
close #255 
- add a more robust computation of directory change
- add a check for "important_files"
- ensure that "preview-webhook" is not filtered by diffs (that feels inappropriate)

we ran into a case where a "subdirectory" of a watched directory changing (i.e. my/path, where we watch my/) would not get built. similarly, changing build infrastructure like our "version script" would not trigger new builds either.

A few open questions:
- should we remove `get-diffs.py` from the "important_files"
- should we consider _excluding_ unimportant files rather than _including_ important? i.e. should we err on the side of building too much (if a new file / directory is added) or too little (if a new global / CI affecting file is added and not added to "important_files")
- is there a better way to think about this problem?
- should we remove the "build protection" (build pass required to merge) since some builds may not run, now? Maybe a way to have an "outcome" build that succeeds only if all executed builds in the matrix pass, which we can require?
- should there be a way to ensure that "all builds get run"? Not sure what this would look like... a slash command or something?

TODO: ensure that builds will run on main when they land...

### DIFF
--- a/.github/workflows/build-preview.yaml
+++ b/.github/workflows/build-preview.yaml
@@ -18,7 +18,13 @@ jobs:
     outputs:
       matrix: ${{ steps.set-matrix.outputs.matrix }}
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-python@v3
+        with:
+          python-version: '3.9'
+          architecture: 'x64'
       - id: set-matrix
         run: |
           # filter the matrix configuration
@@ -28,7 +34,9 @@ jobs:
           # NOTE: blank which input here means that every product is selected
           MATRIX_FILTER=$(echo "$MATRIX" | jq -Mcr 'map(select((.product | startswith("${{ github.event.inputs.which }}"))or("all"=="${{ github.event.inputs.which }}")))')
           echo "Filtered Matrix: $MATRIX_FILTER"
-          echo "::set-output name=matrix::$MATRIX_FILTER"
+          MATRIX_DIFF_FILTER=$(echo "$MATRIX_FILTER" | ./get-diffs.py -i -t origin/main)
+          echo "Filtered Diff Matrix: $MATRIX_DIFF_FILTER"
+          echo "::set-output name=matrix::$MATRIX_DIFF_FILTER"
 
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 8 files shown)


.github/workflows/build-preview.yaml
```diff
@@ -42,7 +50,15 @@ jobs:
...
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
+
+      # setup python
+      - uses: actions/setup-python@v3
+        with:
+          python-version: '3.9'
+          architecture: 'x64'
+          cache: 'pip'
+      - run: pip install -r requirements.txt
```
bschwedler, line 61: Could we switch over to using `urillib` instead of `requests` so we can just use the standard library?

colearendt:
@aronatkins that's a great call. That is _certainly_ much cleaner 😅 I was mostly cribbing from a pattern that helm uses, and a bit ignorant of niceties such as the one you mention. A few tricks / challenges that I haven't thought through yet:

- a way to reference not `origin/main`, but the previous parent commit (i.e. when running on main). Probably easy enough with a shell command
- where we do the check. Are we altering the matrix so that fewer jobs actually get launched as we do here, or are we checking at runtime (i.e. launch all the jobs and see whether they should exit immediately or not). I like filtering the matrix, which we could potentially do using this mechanism within the python loop
- escape hatches for "build everything" (i.e. in the daily cron job for preview, when CI-driving files change, etc.)

I do like how elegant that is though 😄 I will revisit and see if I can simplify

.github/workflows/build-preview.yaml
bschwedler: python2 had `urllib`, `urllib2`, and `urllib3` as part of the standard library.  `requests` has an a nicer interface with which to interact, but since we are using it minimally it should be easy to replace with `urllib` in py3.

colearendt:
For the record, since this stagnated. The simple solution was to disable "rebase commits." Squash commits are safe b/c the previous commit was on main, and merge commits are safe b/c the previous commit was on main. Still much work to do teasing through these edge cases 😄 

I also am hoping to take a look at the community to see if/how other teams have resolved this issue.

bschwedler:
Closing this out since the PR is based on an older workflow that did not use `docker buildx bake`.

We still want to revisit the build frequency of these images, which will likely be addressed by the build tooling refactor we are currently working through.

PR #317, opened 2022-05-24
add custom azure-ml license manager to a dedicated path

colearendt:
unpack the dedicated monitor build to /opt/rstudio/bin, configure
rserver.conf and statup.sh to point at the new binary


bdeitte:
How about matching the path we use for the marketplaces?  Other than that, looks good to me.  How the marketplace is setup:

echo 'server-license-manager-path=/opt/rstudio-license/license-manager'>> /etc/rstudio/rserver.conf
mkdir -p /opt/rstudio-license
mv /tmp/license-manager /opt/rstudio-license/license-manager
chown root.root /opt/rstudio-license/license-manager
chmod 0755 /opt/rstudio-license/license-manager

colearendt:
Fixed up for consistency! Thanks Brian!

blairj09:
Thanks Cole!

PR #321, opened 2022-05-25
move sed commands for update-version to Justfile

colearendt:
quoting, shelling out to sed, and getting sed parameters to function dynamically within `make` was a nightmare. As a result, I am attempting a switch to `just`, at least for this target. `make` calls `just` here, hopefully allowing for backwards compatibility (provided that you have `just` installed 😅 )

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ data
 *.lic
 .idea/
 tmp*/
+.python-version
--- a/Justfile
+++ b/Justfile
@@ -1,0 +1,66 @@
+set positional-arguments
+
+vars := "-i ''"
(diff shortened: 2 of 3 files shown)


aronatkins (commented):
I often use standalone `bash` scripts when `Makefile` and `justfile` targets grow too complicated. It can be helpful to only have one set of variables / expansion rules in play. 😁

colearendt:
Yes! So true 😅 Although I must admit, I was struggling with the bashism to get `sed` args to be passed in properly. (they kept getting interpreted as `sed -i " ' '"` 😱 )

PR #331, opened 2022-06-02
Build `dev` and `dev-rspm` daily during CRON

tylfin:


colearendt (changes requested):
Thanks for the PR!! A couple of things:

- This will essentially cause _every_ PR / push to run this workflow 3 times, so will cause a huge forking of work / builds. Not just the schedule / cron
- We would also need to make changes elsewhere in this workflow definition to handle branching https://github.com/rstudio/rstudio-docker-products/blob/ee254e182cbbd9a6668ab84b67749004127564fa/.github/workflows/build-preview.yaml#L59-L62

I'm inclined to wait for a formal GHA feature for this to integrate it into CRON: https://github.community/t/scheduled-builds-of-non-default-branch/16306/7

In the meantime, I am inclined to use the preview webhook and run a cron on connect.rstudioservices.com ([example app here that I could easily modify for this purpose](https://connect.rstudioservices.com/connect/#/apps/18b12d82-aea0-4972-98c7-545ef30ddcfa/access/648)). It's gross and hacky, but saves us a ton of builds. The other thing we could do is fork this cron out from the "on push" action, which is a little less than ideal, but maybe simpler (I have used `vimdiff` in the past to keep things in sync / as close as possible)

Thoughts on the best approach?

tylfin:
@colearendt Ah good point! How does this look?

PR #333, opened 2022-06-15
dev - do not merge into main - install the jupyterlab extension

melissa-barca:
Now that jupyterlab_rsw is on PyPI we can use `pip` to install the most recent build. 
This PR is just to merge the change in to the dev image for now, but when 2022.06 is released, we should merge this into main.


atheriel (approved):
This looks fine to me, provided the test failure is unrelated this can probably be merged.

PR #337, opened 2022-06-28
Enable "entrypoint.d" for Connect image

rstub:
Uses the ideas at https://www.camptocamp.com/en/news-events/flexible-docker-entrypoints-scripts as starting point.

By adding new scripts to `/docker-entrypoint.d/` one can add additional steps to the start-up process. And if one runs the image with `CMD=bash` one gets into a state where all these steps have been applied.

### DIFF
--- a/connect/docker-entrypoint.d/50-activate-license.sh
+++ b/connect/docker-entrypoint.d/50-activate-license.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+set -e
+set -x
+
+# Activate License
+RSC_LICENSE_FILE_PATH=${RSC_LICENSE_FILE_PATH:-/etc/rstudio-connect/license.lic}
+if ! [ -z "$RSC_LICENSE" ]; then
+    /opt/rstudio-connect/bin/license-manager activate $RSC_LICENSE
+elif ! [ -z "$RSC_LICENSE_SERVER" ]; then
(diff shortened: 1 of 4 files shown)


connect/docker-entrypoint.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+DIR=/docker-entrypoint.d
+if [[ -d "$DIR" ]]; then
+  /bin/run-parts --verbose "$DIR"
```
aronatkins, line 5: What behavior do we want if any part errs?

connect/docker-entrypoint.d/50-activate-license.sh
```diff
@@ -0,0 +1,18 @@
...
+if ! [ -z "$RSC_LICENSE" ]; then
+    /opt/rstudio-connect/bin/license-manager activate $RSC_LICENSE
+elif ! [ -z "$RSC_LICENSE_SERVER" ]; then
+    /opt/rstudio-connect/bin/license-manager license-server $RSC_LICENSE_SERVER
+elif test -f "$RSC_LICENSE_FILE_PATH"; then
+    /opt/rstudio-connect/bin/license-manager activate-file $RSC_LICENSE_FILE_PATH
+fi
+
+# ensure these cannot be inherited by child processes
+unset RSC_LICENSE
```
aronatkins, line 17: Does this still modify the inherited environment?

PR #339, opened 2022-06-29
Update RStudio Connect to version 2022.06.0

Lytol:
Note that `make update-versions` also bumped `ARG RSW_VERSION=2022.02.3+492.pro3` in `helper/workbench-for-microsoft-azure-ml/Dockerfile".  Not sure if that was an oversight that is now rectified or if something funkier is happening...


colearendt (approved):
LGTM! Thanks for running that. I think that just means it was missed when RSW upgraded to the most recent patch release 😅

PR #336, opened 2022-06-24
Rename workbench jupyterlab extension - dev branch only

melissa-barca:
Rename the Workbench JupyterLab extension on the dev branch once https://github.com/rstudio/rstudio-pro/pull/3593 has been merged and a new build has been uploaded to PyPI. The build will fail on this branch until that has been done.

This should not be merged into main until RSW 2022.07.0 is released.

PR #297, opened 2022-04-23
Workbench updates for 2022.07 release

colearendt:
Adding quarto, jupyter, and jupyterlab features for the 2022.07 release

PR #345, opened 2022-07-12
install jupyterlab extension

edavidaja:


colearendt (commented):
Do we want to add this to session images too? Should we close this in favor of those other PRs to get quarto too? (#336  and #297 )

edavidaja:
Yep, didn't see those other PRs; let's move those forward.

PR #346, opened 2022-07-12
Refactor Workflows & Add Jammy Images

dkasabovn:
#### Intent

Previous workflows included logic within the yaml in order to tag and push various versions. I wanted to keep the steps between workflows the same while retaining log visibility.

#### Approach

Rather than a `Makefile` I have included a `Justfile` with one main command: `build`. Build takes in 4 arguments and then builds and tags images correctly based upon the arguments. This allows all of the logic that used to be in the workflow to migrate over to the `Justfile` making things easier to read and debug.

**Docker Subfolder**
The addition of the docker subfolder was somewhat arbitrary. I didn't want to have Dockerfiles directly underneath package folders and wanted to keep them easy to locate. I don't have strong opinions on this, so if anything else is suggested I am ok to switch.

**Makefile**
This PR grew in scope a lot since starting it. The makefile is a remnant of a pr where I was just to add Jammy images and build them. I hesitated on removing the Makefile entirely, as I'm not sure what the plans are here. Do we remove the makefile? or keep it?

**Justfile Ugliness**
This line is probably the biggest embodiment of the jankiness of using Just
```
-build-arg "${short_name}"_VERSION=$version  ${rsw_download_url_arg} --file=./{{PRODUCT}}/docker/{{OS}}/Dockerfile {{PRODUCT}}
```
We have Just variable substitution with `{{PRODUCT}}`, bash substitution with `"${short_name}"`, and then variables that may be empty or may have arguments.

This was needed as I felt it made sense for build-preview to be encapsulated in one command rather than 3 for each package (build-preview-rsw). As a consequence there are a lot of if statements needed in order to select the right options for builds.

#### Automated Tests

Testing with [ACT](https://github.com/nektos/act) is not ideal. Things like docker-compose not being available by default make it necessary to add additional steps into workflows when performing local testing as compared to production. This would mean testing, removing additions made for testing, then pushing everytime a change needs to be made.

#### Notes

**Why is Jammy here?**
Adding Jammy images was the original goal of this PR before it evolved to a Workflow refactor. However, I think its important to keep Jammy changes bundled with the workflow refactor as it demonstrates how the workflows will support different build targets. I can always split it up, but I don't think that the Jammy changes include much more than a couple of one-line edits from the bionic dockerfiles.

**Just As a Whole**
As Just was used by other repos, I also used Just. However, I'm not sure that Just is ideal. It seems to me that calling a shell/python executable from Make is almost a cleaner solution than using a Justfile. Doing so would prevent the chimeric appearance of the `build-preview` and `build-release` commands that can only be understood by someone who knows bash **and** Just. I'm just not convinced that Just is *the* solution to situations where Makefiles need a bit more logic.


.github/workflows/build-preview-v2.yaml
```diff
@@ -0,0 +1,94 @@
...
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.BUILD_PAT }}
+
+      - name: Push Images (just)
+        id: push-images
+        run: |
+          just p ${{ steps.build-images.outputs.tags }}
+
```
colearendt, line 94: Beware missing newline at the end of the file. Usually you can ensure this with a setting in your editor.
image

.github/workflows/build-preview-v2.yaml
```diff
@@ -0,0 +1,94 @@
...
+        with:
+          path: /tmp/.buildx-cache
+          key: ${{ runner.os }}-buildx-${{ github.sha }}
+          restore-keys: |
+            ${{ runner.os }}-buildx-
+
+      - name: Build Images (just)
+        id: build-images
+        run: |
+          TAGS=`just b ${{ matrix.config.type }} ${{ matrix.config.package }} ${{ matrix.config.os }} ${{ steps.buildx.outputs.name }}`
```
colearendt, line 59: I personally am more in favor of using the "full" names for the just targets (i.e. "build" instead of "b"). It's not a whole lot more writing, and it makes the command much more natural to read and understand (rather than having to go read the name of the step, etc.). 

The aliases are nice as a convenience interactively, but in code the more clear name should be worth the extra characters

Justfile
```diff
@@ -0,0 +1,119 @@
...
+
+    echo rstudio/rstudio-{{PRODUCT}}:{{OS}}-latest \
+        rstudio/rstudio-{{PRODUCT}}:{{OS}}-"${safe_version}" \
+        rstudio/rstudio-{{PRODUCT}}:{{OS}}-"${safe_version}"--{{SHA_SHORT}} \
+        ghcr.io/rstudio/rstudio-{{PRODUCT}}:{{OS}}-latest \
+        ghcr.io/rstudio/rstudio-{{PRODUCT}}:{{OS}}-"${safe_version}" \
+        ghcr.io/rstudio/rstudio-{{PRODUCT}}:{{OS}}-"${safe_version}"--{{SHA_SHORT}} 
+    
+
+build-preview $TYPE $PRODUCT OS VERSION="" BRANCH=`git branch --show`:
```
colearendt, line 60: This is pretty cool!

Justfile
```diff
@@ -0,0 +1,119 @@
...
+        buildx_args="--cache-from=type=local,src=/tmp/.buildx-cache --cache-to=type=local,dest=/tmp/.buildx-cache"
+    fi
+
+    docker buildx --builder=$BUILDX build --load $buildx_args -t rstudio/rstudio-{{PRODUCT}}-preview:"${branch_prefix}"{{OS}}-"${safe_version}" \
+        -t rstudio/rstudio-{{PRODUCT}}-preview:"${branch_prefix}"{{OS}}-{{TYPE}} \
+        -t ghcr.io/rstudio/rstudio-{{PRODUCT}}-preview:"${branch_prefix}"{{OS}}-"${safe_version}" \
+        -t ghcr.io/rstudio/rstudio-{{PRODUCT}}-preview:"${branch_prefix}"{{OS}}-{{TYPE}} \
+         --build-arg "${short_name}"_VERSION=$version ${rsw_download_url_arg} --file=./{{PRODUCT}}/docker/{{OS}}/Dockerfile {{PRODUCT}}
+
+    echo rstudio/rstudio-{{PRODUCT}}-preview:"${branch_prefix}"{{OS}}-"${safe_version}" \
```
colearendt, line 93: Would you mind saying a bit in the PR or a comment about the variable syntax and why you ended up needing to go this direction?

Makefile
```diff
@@ -77,15 +80,64 @@ update-versions:  ## Update the version files for all products
...
+build-all: $(BUILD_PACKAGES)
+
+test-all: $(TEST_PACKAGES)
+
+test-all-i: $(TEST_PACKAGES_I)
 
 rsw: workbench
 workbench:  ## Build Workbench image
-	docker build -t rstudio/rstudio-workbench:$(RSW_TAG_VERSION) --build-arg R_VERSION=$(R_VERSION) --build-arg RSW_VERSION=$(RSW_VERSION) workbench
+	docker build -t rstudio/rstudio-workbench:$(RSW_TAG_VERSION) --build-arg R_VERSION=$(R_VERSION) --build-arg RSW_VERSION=$(RSW_VERSION) workbench --file=./workbench/docker/bionic/Dockerfile
```
colearendt, line 135: Were these the start of makefile changes that you ended up stopping progress on?

Would be good to have context - I would have expected other builds to go through a similar refactor

Justfile
```diff
@@ -0,0 +1,119 @@
...
+    do
+        docker push "${image}"
+    done
+    
+
+test-image $TYPE $PRODUCT +IMAGES:
+    #!/usr/bin/env bash
+    set -euxo pipefail
+    images="{{IMAGES}}"
+    version=`just gv $PRODUCT --type=$TYPE --local`
```
colearendt, line 111: Similar for the justfile - expounding that `gv` = `getversion` can make this read a lot clearer

colearendt (changes requested):
Thanks! If you could say a bit more in your PR about context and choices, I think that would be super helpful. For instance:

- Why create the `docker/` subfolder? It feels redundant? Did you use this pattern from somewhere else? Is it preferable for some reason? 
- A note about things that are included but not actively in use (i.e. the Makefile changes, actions/single/action.yml), etc.
- Notes about any particularly complex problems you solved (can help reviewers pay more close attention there and understand your reasoning for certain choices)
- what is "ACT" testing?

Context helps a lot for someone not as close to the problem (me), especially when reviewing a sizable PR like this one 😄 

I still need to dig in a bit further and poke around at things!

.github/workflows/build-preview-v2.yaml
atheriel: There are a few more instances of this below that could be fixed, too.

.github/workflows/build-preview-webhook-v2.yaml
```diff
@@ -0,0 +1,101 @@
...
+        id: test-images
+        env:
+          RSC_LICENSE: ${{ secrets.RSC_LICENSE }}
+          RSPM_LICENSE: ${{ secrets.RSPM_LICENSE }}
+          RSW_LICENSE: ${{ secrets.RSW_LICENSE }}
+        run: |
+          just t ${{ matrix.config.type }} ${{ matrix.config.package }} ${{ steps.build-images.outputs.tags }}
+
+      - name: Login to Docker Hub
+        if: ${{ github.ref == 'refs/heads/main' || github.ref == 'refs/heads/dev' || github.ref == 'refs/heads/dev-rspm' }}
```
atheriel, line 84: Since this is a webhook-based workflow, we should probably skip the branch check here and below.

.github/workflows/build-release-v2.yaml
```diff
@@ -0,0 +1,88 @@
...
+        id: test-images
+        env:
+          RSC_LICENSE: ${{ secrets.RSC_LICENSE }}
+          RSPM_LICENSE: ${{ secrets.RSPM_LICENSE }}
+          RSW_LICENSE: ${{ secrets.RSW_LICENSE }}
+        run: |
+          just t ${{ matrix.config.type }} ${{ matrix.config.package }} ${{ steps.build-images.outputs.tags }}
+
+      - name: Login to Docker Hub
+        if: ${{ github.ref == 'refs/heads/main' || github.ref == 'refs/heads/dev' || github.ref == 'refs/heads/dev-rspm' }}
```
atheriel, line 71: I think this is obviated by the branch filter above, and can be removed here and below.

.github/workflows/build-release-v2.yaml
```diff
@@ -0,0 +1,88 @@
...
+
+      - name: Login to ghcr.io
+        if: ${{ github.ref == 'refs/heads/main' || github.ref == 'refs/heads/dev' || github.ref == 'refs/heads/dev-rspm' }}
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.BUILD_PAT }}
+
+      # - name: Push Images (just)
```
atheriel, line 85: I'm not sure I understand why this is commented out here.

Justfile
```diff
@@ -0,0 +1,119 @@
+#!/usr/bin/env just --justfile
+
+alias b := build
```
atheriel, line 3: As Cole mentioned above, these are fine for local development but I don't think aliases should be used in the CI workflows.

atheriel:
I'm a bit unclear as to why the workflow changes need to land alongside the new Jammy images -- is it possible to break these into separate PRs, or at least into separate commits?

package-manager/docker/jammy/Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+RUN curl -L -o /usr/local/bin/tini https://github.com/krallin/tini/releases/download/v${TINI_VERSION}/tini && \
+    chmod +x /usr/local/bin/tini
+
+COPY --chmod=0775 startup.sh /usr/local/bin/startup.sh
+
+EXPOSE 4242/tcp
+EXPOSE 2112/tcp
+
+# Download RStudio Package Manager ---------------------------------------------#
+ARG RSPM_VERSION=2022.04.0-7
```
atheriel, line 25: Does this RSPM release actually support Jammy? If not, maybe we should wait for the next one (which probably will) before adding this.

workbench/docker/jammy/Dockerfile
```diff
@@ -0,0 +1,160 @@
...
+RUN mkdir -p /var/lib/rstudio-server/monitor/log && \
+    chown -R rstudio-server:rstudio-server /var/lib/rstudio-server/monitor && \
+    mkdir -p /startup/custom/ && \
+    echo '\n# allow home directory creation\nsession required pam_mkhomedir.so skel=/etc/skel umask=0022' >> /etc/pam.d/common-session
+
+EXPOSE 8787/tcp
+EXPOSE 5559/tcp
+
+ENTRYPOINT []
+CMD ["/usr/bin/supervisord", "-c", "/etc/supervisor/supervisord.conf"]
```
atheriel, line 160: Nit: missing newline.

get-matrix.py
```diff
@@ -0,0 +1,52 @@
...
+                if build_type == "preview" and package in ["connect", "package-manager"]:
+                    continue
+                matrix_conf.append({
+                    "package": package,
+                    "os": oss,
+                    "type": build_type
+                })
+    
+    print(f"Generated matrix conf: {json.dumps(matrix_conf)}", file=sys.stderr)
+    print(json.dumps(matrix_conf))
```
atheriel, line 52: Nit: missing newline.

Justfile
```diff
@@ -0,0 +1,119 @@
+#!/usr/bin/env just --justfile
```
atheriel, line 1: It would be nice if the `docker` command here was substitution-friendly so we could build with `nerdctl` or `podman`.

.github/workflows/build-preview-v2.yaml
```diff
@@ -0,0 +1,94 @@
...
+          echo "::set-output name=tags::$TAGS"
+        
+      - name: Show Image Sizes
```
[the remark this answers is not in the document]
dkasabovn: It was in the other workflows, so I just kept it.

.github/workflows/build-release-v2.yaml
dkasabovn: I'm pretty sure removing this `if` would push release on any pull request, let me know if this is intended.

Justfile
dkasabovn: Added context, let me know if it is enough

Makefile
dkasabovn: Added in description

.github/workflows/build-release-v2.yaml
dkasabovn: not wanting to push release versions before everything is approved

Justfile
dkasabovn: Not sure what this means entirely. I've never built with podman or nerdctl, so if you could throw me an example, I can look for a way to integrate this into the build.

.github/workflows/build-preview-v2.yaml
```diff
@@ -0,0 +1,94 @@
...
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: extractions/setup-just@v1
+      
+      - name: Set Matrix
+        id: set-matrix
+        run: |
+          MATRIX=`just gm --types preview daily --os bionic jammy --packages workbench connect package-manager | jq -Mcr`
```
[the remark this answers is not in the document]
dkasabovn: Will look into this in a bit

package-manager/docker/jammy/Dockerfile
dkasabovn: No, but I figured it doesn't hurt to add this in for now. Let me know what you think.

dkasabovn:
> I'm a bit unclear as to why the workflow changes need to land alongside the new Jammy images -- is it possible to break these into separate PRs, or at least into separate commits?

Added context in pr header

.github/workflows/build-preview-v2.yaml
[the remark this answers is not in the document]
colearendt: Yeah, it was meant for debugging and an easy way to see how image size is affected by various PRs / etc. (before we push)

.github/workflows/build-release-v2.yaml
atheriel: Ah, you're right!

connect/docker/jammy/Dockerfile
```diff
@@ -0,0 +1,124 @@
+ARG R_VERSION=3.6.2
```
bschwedler, line 1: Is it possible to use to a more recent `R` Version like `4.1` or `4.2` since these are newly created images?

I don't have the context as to whether each of the products all support the newer R versions.


This would also apply to the other product images.

dkasabovn: I certaintly change that, but we'd have to make a decision image by image on this I believe

PR #340, opened 2022-06-30
Update `dev-rspm` branch for consistent Launcher directory fix

glin:
Updates the `dev-rspm` branch for the consistent Launcher directory fix, https://github.com/rstudio/rstudio-docker-products/pull/335. This wasn't straightforward because https://github.com/rstudio/rstudio-docker-products/pull/335 conflicted with the `dev-rspm` branch (we moved `DataDir` from `/data` to `/var/lib/rstudio-pm`).

What I did was rebase `dev-rspm` against `main` before that PR merged (https://github.com/rstudio/rstudio-docker-products/commit/dc36cdc58d79a151367a147b926ab30f0e4d771d), then merged `main` into a branch based on `dev-rspm`, and manually resolved conflicts, changing `LauncherDir = /data/launcher_internal` to `LauncherDir = /var/lib/rstudio-pm/launcher_internal`. This lets the `dev-rspm` branch merge cleanly into `main` again, but adds some extra merge commits. Alternatively, I could try a complete rebase if that's better, but that would be harder to review on GitHub I think.

I noticed that the `dev-rspm` branch still needs a NEWS item with the root->non-root and data directory changes, but didn't add that here, just an FYI before we merge back to `main`.

Testing is similar to https://github.com/rstudio/rstudio-docker-products/pull/335. Without a persistent data directory, confirm that RSPM starts up and that the `/var/lib/rstudio-pm/launcher_internal` is initialized as the launcher data directory:

```sh
docker run -it --rm \
    -p 4242:4242 \
    -e RSPM_LICENSE=$RSTUDIO_PM_LICENSE \
    --name rspm \
    rstudio/rstudio-package-manager:2022.04.0-7

docker exec rspm ls -la /var/lib/rstudio-pm/launcher_internal
# total 40
# drwxr-xr-x 5 rstudio-pm rstudio-pm 4096 Jun 30 20:27 .
# drwx------ 1 rstudio-pm rstudio-pm 4096 Jun 30 20:27 ..
# drwx------ 3 rstudio-pm rstudio-pm 4096 Jun 30 20:27 custom_logs
# drwx------ 3 rstudio-pm rstudio-pm 4096 Jun 30 20:27 jobs
# -rw------- 1 rstudio-pm rstudio-pm  562 Jun 30 20:27 launcher.conf
# -rw------- 1 rstudio-pm rstudio-pm  228 Jun 30 20:27 launcher.kubernetes.conf
[...]

PR #348, opened 2022-07-20
Bump Package Manager to version 2022.07.0-9 and merge non-root changes

tylfin:
Merging in changes from dev-rspm and bumping image to `2022.07.0-9`


atheriel:
CI seems to be failing for a mysterious reason -- does the image seem to work correctly in your testing?

tylfin:
@atheriel Yup, we've been using the `dev-rspm` branch in our K8s cluster cc @rodrigobrim

package-manager/Dockerfile
```diff
@@ -1,6 +1,6 @@
 ARG R_VERSION=3.6.2
 FROM rstudio/r-base:${R_VERSION}-bionic
-LABEL maintainer="RStudio Docker <docker@rstudio.com>"
+LABEL maintainer="RStudio Package Manager Team <rspm@rstudio.com>"
```
colearendt, line 3: Is this email address publicly routable? (some internal email addresses / groups are not). Moreover, not sure whether this is something y'all have standardized on handling for the other images? We have used `docker@rstudio.com` as an "IT oriented" place for inbound image notices

package-manager/README.md
```diff
@@ -14,8 +14,13 @@ Docker images for RStudio Professional Products
    changes. We
    provide [instructions for building](https://github.com/rstudio/rstudio-docker-products#instructions-for-building) for
    these cases.
+1. The Package Manager image uses the `No Sandbox` option documented
```
colearendt, line 17: Related to this, I suspect we should have an "upgrade notice" or something with the steps needed to upgrade to 2022.07, changing the default RunAs user, etc.

package-manager/README.md
```diff
@@ -34,35 +39,30 @@ Open [http://localhost:4242](http://localhost:4242) to access RStudio Package Ma
...
 
-#### Configuration
+## Configuration
 
 RStudio Package Manager is configured via the`/etc/rstudio-pm/rstudio-pm.gcfg` file. You should mount this file as
 a volume from the host machine. Changes will take effect when the container is restarted.
 
-Be sure the config file has these fields:
+Be sure the config file has the `[HTTP].Listen` field configured. See a complete example of that file at
+`package-manager/rstudio-pm.gcfg`.
```
colearendt, line 59:
```suggestion
[`package-manager/rstudio-pm.gcfg`](./rstudio-pm.gcfg).
```

colearendt (changes requested):
I haven't tried this myself yet, but a handful of thoughts. One thing in particular:

For sizable changes like this, we should write up a thorough overview of changes in NEWS.md so users have a jumping off point of "where to dig further" if they are running, upgrading, or extending this image. Namely:

- run as user changes
- no longer need sandboxing / the privileged flag
- disabled metrics by default
- Changed DataDir from `/data` to `/var/lib/rstudio-pm`

colearendt:
Oh, also - y'all missed updating the Makefile (which currently drives the CI build version): https://github.com/rstudio/rstudio-docker-products/blob/main/RELEASE.md

Worth noting that `make update-versions` fails on certain OS's... 😓 Hence #321

colearendt:
And re: CI, the container is failing to start on this:
```
2022/07/20 16:27:40 Creating pypi directory: /var/lib/rstudio-pm/pypi
2022/07/20 16:27:40 Creating launcher directory: /data/launcher_internal
2022/07/20 16:27:40 Error: Unable to initialize the configuration: mkdir /data: permission denied
```

From here in the config:
```
LauncherDir = /data/launcher_internal
```

package-manager/Dockerfile
tylfin:
```suggestion
LABEL maintainer="RStudio Docker <docker@rstudio.com>"
```

package-manager/README.md
tylfin: Would this go in the [NEWS](https://github.com/rstudio/rstudio-docker-products/blob/main/package-manager/NEWS.md) or elsewhere?

package-manager/rstudio-pm-float.gcfg
```diff
@@ -18,7 +18,7 @@ RVersion = /opt/R/3.6.2/
 ;DataDir = /mnt/rspm/data
 ; Use a consistent location for the Launcher directory. The default location
 ; is based on the hostname, and the hostname may be different in each container.
-LauncherDir = /data/launcher_internal
+LauncherDir = /var/lib/rstudio-pm/launcher_internal
```
glin, line 21: I made this change in the `dev-rspm` branch at https://github.com/rstudio/rstudio-docker-products/pull/340, so it should've been included already, hmm. Is this PR branch made off the latest `dev-rspm` changes? It would probably be better to use the same commit from https://github.com/rstudio/rstudio-docker-products/pull/340 to get the other changes from that PR and avoid conflicts between `main` and `dev-rspm`.

package-manager/README.md
colearendt: NEWS at a minimum IMO. It also may be worth a note / pointer in the README? 🤷 Up to y'all!

package-manager/Dockerfile
```diff
@@ -16,13 +16,8 @@ ARG TINI_VERSION=0.18.0
 RUN curl -L -o /usr/local/bin/tini https://github.com/krallin/tini/releases/download/v${TINI_VERSION}/tini && \
     chmod +x /usr/local/bin/tini
 
-COPY --chmod=0775 startup.sh /usr/local/bin/startup.sh
-
-EXPOSE 4242/tcp
-EXPOSE 2112/tcp
```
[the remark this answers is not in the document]
atheriel: I recall that Tyler had a reasonable answer in the original PR, we could dig it up.

package-manager/NEWS.md
```diff
@@ -1,3 +1,13 @@
+# 2022-07-21
+
+- This update introduces a breaking change that sets the default user to `rstudio-pm` instead of root. This
+  may require permission changes for existing environments.
+- The `Server.DataDir` setting has been set back to default, for more information see the
```
colearendt, line 5: Can we tell what these are explicitly and how to resolve? We know that it used to be `/data` and right now the default is `/var/lib/rstudio-pm`, right? I think that would be useful data. Something akin to:

To mirror old behavior, set
```
[Server]
DataDir = /data
```

It'd also be good to have a "BREAKING: " note in front of these bullets where a change is potentially breaking (helps admins see quickly what they need to change).

package-manager/README.md
```diff
@@ -4,18 +4,25 @@ Docker images for RStudio Professional Products
...
 **IMPORTANT:** There are a few things you need to know before using these images:
 
-1. These images are provided as a convenience to RStudio customers and are not formally supported by RStudio. If you
+1. This image is still underdevelopment and new versions may introduce **BREAKING** changes, read through the
+   [NEWS](./NEWS.md) to understand these changes before updating.
+2. These images are provided as a convenience to RStudio customers and are not formally supported by RStudio. If you
    have questions about these images, you can ask them in the issues in the repository or to your support
    representative, who will route them appropriately.
-1. Outdated images will be removed periodically from DockerHub as product version updates are made. Please make plans to
+3. Outdated images will be removed periodically from DockerHub as product version updates are made. Please make plans to
```
colearendt, line 12: Worth noting that GitHub .md syntax handles this numbering automatically (hence the all 1s, and it displays fine 😄 )

package-manager/README.md
```diff
@@ -4,18 +4,25 @@ Docker images for RStudio Professional Products
 
 **IMPORTANT:** There are a few things you need to know before using these images:
 
-1. These images are provided as a convenience to RStudio customers and are not formally supported by RStudio. If you
+1. This image is still underdevelopment and new versions may introduce **BREAKING** changes, read through the
```
colearendt, line 7: I personally don't think we should make too much of connecting these two (under development and breaking changes). The breaking changes we have here are completely reversible with configuration, and we should have license to make BREAKING changes when it is for the good of our customers IMO 🤷 The TL; DR; is definitely "read the NEWS" 😄

colearendt (commented):
Looking awesome! Just a few nits on wording 😄

colearendt (approved):
LGTM!

Issue #349, opened 2022-07-20
Remove privileged containers requirement


kfeinauer:
The documentation currently states that all product containers need to be run with the `--privileged` Docker flag, due to historical sandboxing issues. However, this is no longer the case as all products can run in unprivileged Docker containers.

- [x] Remove documentation for `--privileged` requirement
- [x] Update RSW launcher.conf (https://github.com/rstudio/rstudio-docker-products/blob/main/workbench/conf/launcher.conf) to ensure `unprivileged-mode=1` is set to allow unprivileged Docker execution
- [x] Smoke test all products to ensure proper functionality without `--privileged` flag.

PR #353, opened 2022-07-21
#349 Remove privileged containers requirement

ianpittwood:
- Removed `--privileged` flag for all documentation and definitions/scripts except for Connect.
- Added NEWS entry for change with link to Privileged Requirements support article.

### DIFF
--- a/connect-content-init/docker-compose.test.yml
+++ b/connect-content-init/docker-compose.test.yml
@@ -5,7 +5,6 @@ services:
     image: $IMAGE_NAME
     command: /run_tests.sh
     entrypoint: []
-    privileged: true
     environment:
       # uses .env by default
       - RSC_VERSION
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,7 +8,6 @@ services:
       args:
         RSW_VERSION: 2022.07.0+548.pro5
     image: rstudio/rstudio-workbench:2022.07.0-548.pro5
-    privileged: true
     environment:
       RSW_LICENSE: ${RSW_LICENSE}
       LICENSE_SERVER: ${RSW_LICENSE_SERVER}
@@ -43,7 +42,6 @@ services:
       args:
         RSPM_VERSION: 2022.07.0-9
     image: rstudio/rstudio-package-manager:2022.07.0-9
-    privileged: true
     environment:
       RSPM_LICENSE: ${RSPM_LICENSE}
       LICENSE_SERVER: ${RSPM_LICENSE_SERVER}
--- a/helper/float/README.md
+++ b/helper/float/README.md
@@ -35,7 +35,7 @@ export RSC_LICENSE_SERVER=rsc-float-lic:8999
 export RSPM_LICENSE_SERVER=rspm-float-lic:8969
 
 # RStudio Workbench
-docker run --privileged -it \
+docker run -it \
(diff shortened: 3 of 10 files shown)


docker-compose.yml
```diff
@@ -26,7 +25,6 @@ services:
       args:
         RSC_VERSION: 2022.06.2
     image: rstudio/rstudio-connect:2022.06.2
-    privileged: true
```
colearendt, line 29: I think this one is still required?

connect-content-init/docker-compose.test.yml
```diff
@@ -5,7 +5,6 @@ services:
     image: $IMAGE_NAME
     command: /run_tests.sh
     entrypoint: []
-    privileged: true
```
colearendt, line 8: Good catch!

package-manager/README.md
```diff
@@ -109,14 +108,14 @@ to explore using a license server, license file, or custom image with manual int
 export RSPM_LICENSE=XXXX-XXXX-XXXX-XXXX-XXXX-XXXX-XXXX
 
 # Run without persistent data and using an external configuration
-docker run -it --privileged \
```
colearendt, line 112: These are about to be removed by the RSPM team landing a bunch of changes in #348 - probably best to leave off changes to the package-manager side, TBH

workbench/README.md
```diff
@@ -29,7 +29,7 @@ To verify basic functionality as a first step:
 export RSW_LICENSE=XXXX-XXXX-XXXX-XXXX-XXXX-XXXX-XXXX
 
 # Run without persistent data using default configuration
-docker run --privileged -it \
```
colearendt, line 32: Just worth noting that workbench needs the other config changes to work in this setup

helper/float/README.md
```diff
@@ -35,7 +35,7 @@ export RSC_LICENSE_SERVER=rsc-float-lic:8999
 export RSPM_LICENSE_SERVER=rspm-float-lic:8969
 
 # RStudio Workbench
-docker run --privileged -it \
```
colearendt, line 38: Nice! Would be good to test these to be sure they never needed `--privileged`

colearendt (commented):
Draft is looking great! Just a few points to keep in mind

docker-compose.yml
ianpittwood: Yes, thank you!

workbench/README.md
ianpittwood: I added launcher.local.conf, are there additional changes that are needed?

package-manager/README.md
ianpittwood: I rebased after their PR merged so I think we should be good now, right?

helper/float/README.md
ianpittwood: I tested the float containers and they seem to be working OK.

tylfin (approved):

colearendt (commented):
LGTM! Only nit is that a NEWS.md entry in workbench probably makes sense since we changed the default configuration for the container

ianpittwood:
@colearendt I added an entry in workbench/NEWS.md per your feedback. Let me know if you'd like different wording!

colearendt (approved):
LGTM!

PR #358, opened 2022-07-26
Refactor folder structure

dkasabovn:
#### Changes
- Flatten product folders
- Uniquely identify `Dockerfile`s by adding the target os as a suffix. Ex: `Dockerfile.bionic`
- Split matrix os key into `os` and `arch`. Ex: `os: bionic/amd64` -> `os: bionic, arch: amd64`

#### Motivation
The current folder structure is not consistent. `r-session-complete` which builds to `bionic` and `centos7` has subfolders for each os while workbench which is currently only built for `bionic` does not have any subfolders. In order to add more operating systems and improve builds it is a good idea to make all folders consistent in structure.

I propose that each package should look like this:
```
package
|   Dockerfile.<os>
|---test
|   |   goss.yaml
|   |   run_tests.sh
|   docker-compose.test.yaml
|   .env
|   <any other files/folders required to build individual package>
```

#### Exceptions
`content` is unique in that it has its own build and matrix configuration. I did not change the folder structure of content at all as making modifications to content would have to be a much larger change than currently intended.

`float` is not currently built in any workflow so I did not modify it.

#### Testing
I used nektos/act for testing, however this is unreliable as act is not 1:1 with GH Actions.


colearendt (commented):
This looks really good! Well done!

A few notes:
- let's wait to merge this early next week so we can make sure we test the new images thoroughly as they get pushed out 😄 
- I suspect this will break some things in the `Makefile` (i.e. local testing, IMAGE_NAMEs, etc.). I think that is fine and we can clean that up as a part of moving to `just` / cleaning up other CI stuff, etc.

dkasabovn:
> This looks really good! Well done!
> 
> A few notes:
> 
> * let's wait to merge this early next week so we can make sure we test the new images thoroughly as they get pushed out 😄
> * I suspect this will break some things in the `Makefile` (i.e. local testing, IMAGE_NAMEs, etc.). I think that is fine and we can clean that up as a part of moving to `just` / cleaning up other CI stuff, etc.
> 
> Oops - one merge conflict in `float` too.
> 
> Also, it's worth noting that this will bring images that _had_ been a smidge hidden (i.e. `float`, `azure`, etc.) into the limelight a bit more. I like the consistency, just not sure if it becomes "too many" top level directories, especially ones that could lead customers astray. I.e. the `float` images should not really be used by clients, and the `azure` image is very rarely what people want who are looking at this repo.

I can 'hide' float as it isn't part of the build pipeline, but I think `workbench-for-microsoft-azure-ml` is a sufficiently descriptive enough name to differentiate itself from `workbench`. At the end of the day I think taking away a bit of logic from the Justfile and making the structure consistent is a good tradeoff, but let me know your thoughts.

Also as far as the makefile I'll probably remove it as I don't think its used.

colearendt (approved):
LGTM! Thanks for handling the ugly merge conflict with that version update script 🙈

PR #362, opened 2022-07-29
Update get-version.py to use elsbeth-geranium index.json

GCRev:
Updates get-version script to use the latest index.json for elsbeth-geranium builds.


colearendt (approved):
LGTM! Thanks!

Issue #350, opened 2022-07-21
Fix azure-ml clamav scanning


kfeinauer:
The azure-ml image build is frequently failing to perform clamav scanning:

```
#35 [clamav 3/4] RUN touch /clamupdate.log &&     chown clamav:clamav /clamupdate.log &&     freshclam --log=/clamupdate.log
#35 0.093 ERROR: Problem with internal logger (UpdateLogFile = /clamupdate.log).
#35 0.094 ERROR: initialize: libfreshclam init failed.
#35 0.094 ERROR: Initialization error!
#35 0.095 ERROR: Can't open /clamupdate.log in append mode (check permissions!).
#35 ERROR: process "/bin/sh -c touch /clamupdate.log &&     chown clamav:clamav /clamupdate.log &&     freshclam --log=/clamupdate.log" did not complete successfully: exit code: 2
```

ianpittwood:
As far as I can tell, this last happened on this build 9 days ago: https://github.com/rstudio/rstudio-docker-products/actions/runs/2734854690

The error itself has been noted as early as 2005 so it seems to be common occurrence. The problem for us is that it seems to happen randomly, making it the worst type of bug. The most [recent discussion](https://ubuntuforums.org/showthread.php?t=2436199) I found noted that the logs are by default written with the ownership `clamav:adm`. I'm currently attempting to modify the way we're logging freshclam and clamscan to be closer to the default methods. Unfortunately, I have not been able to replicate the bug locally though.

bdeitte:
@colearendt any thoughts on this?

bdeitte:
After further discussion, Ian mentioned that he has a PR to switch to build things in a subdir which should always have the right permissions.

PR #366, opened 2022-08-03
Improve available commands around Azure Workbench image

ianpittwood:
While working on #350, I noticed we were missing some of the common commands around build and test for the Azure Workbench image so I implemented them. I also changed the naming to the convention we've been using for other images.

I also added a "DO NOT USE IN PRODUCTION" note for `run-float` so it's more visible to end users.


colearendt (approved):
LGTM!

PR #367, opened 2022-08-04
Use clamav default directory for logging

ianpittwood:
Unfortunately, I was not able to reproduce the permissions issue seen in #350, but writing to the `clamav` default logging directory (created on install) should ensure we don’t have permissions issues for log files. I confirmed that this change works and that the ownership of `/var/log/clamav` is `clamav:adm`.


atheriel (approved):
Makes sense to me.

PR #368, opened 2022-08-08
Refactor Preview Workflow

dkasabovn:
### Changes
Add these build targets to Justfile:
- `build-preview`
- `_rsw-download-url`
- `push-images`
- `test-image`
- `get-version`
- `get-safe-version`

Modify `build-preview.yaml` to use new Justfile targets.

### Motivation
Our current GH Actions contain a lot of logic for getting product versions, generating image tags, and etc. This is not ideal as this logic is duplicated and slightly modified among `build-preview`, `build-preview-webhook`, and `build-latest`. Moving this logic to the `Justfile` will prevent code duplication and make future maintenance easier.

### Testing
If you use [act](https://github.com/nektos/act) make sure you choose the largest image, otherwise docker-compose will not work. *this is a 20 gb image*

### Notes
`BUILDX_PATH` was added as a global Justfile variable because it removes the need for a positional argument within `build-preview` and can be useful locally to anyone who wants to store their `BUILDX_PATH`. If it is empty it is simply ignored.

`build-preview` outputs a list of images built to stdout. These tags will later be used by `push-images` and `test-image`.

### DIFF
--- a/.github/workflows/build-preview.yaml
+++ b/.github/workflows/build-preview.yaml
@@ -13,59 +13,35 @@ on:
 name: build/test/push (preview)
 jobs:
 
-  matrix:
-    runs-on: ubuntu-latest
-    outputs:
-      matrix: ${{ steps.set-matrix.outputs.matrix }}
-    steps:
-      - uses: actions/checkout@v2
-      - id: set-matrix
-        run: |
-          # filter the matrix configuration
-          MATRIX=$(jq -Mcr < matrix-preview.json)
-          echo "Raw Matrix: $MATRIX"
-          echo "Input - which: ${{ github.event.inputs.which }}"
-          # NOTE: blank which input here means that every product is selected
-          MATRIX_FILTER=$(echo "$MATRIX" | jq -Mcr 'map(select((.product | startswith("${{ github.event.inputs.which }}"))or("all"=="${{ github.event.inputs.which }}")))')
-          echo "Filtered Matrix: $MATRIX_FILTER"
-          echo "::set-output name=matrix::$MATRIX_FILTER"
 
   build:
     runs-on: ubuntu-latest
-    name: build-${{ matrix.config.tag_prefix }}${{ matrix.config.product }}:${{ matrix.config.version }}
-    needs: matrix
+    name: build-${{ matrix.config.product }}:${{ matrix.config.type }}:${{ matrix.config.os }}
 
     strategy:
       fail-fast: false
       matrix:
-        config: ${{ fromJson(needs.matrix.outputs.matrix) }}
+        config:
+         - {product: "workbench", type: "daily", os: "bionic"}
+         - {product: "workbench", type: "preview", os: "bionic"}
+         - {product: "connect", type: "daily", os: "bionic"}
+         - {product: "connect-content-init", type: "daily", os: "bionic"}
+         - {product: "package-manager", type: "daily", os: "bionic"}
+         - {product: "r-session-complete", type: "daily", os: "bionic"}
+         - {product: "r-session-complete", type: "daily", os: "centos7"}
+         - {product: "r-session-complete", type: "preview", os: "bionic"}
+         - {product: "r-session-complete", type: "preview", os: "centos7"}
 
     steps:
       - name: Check Out Repo
         uses: actions/checkout@v2
 
+      - name: Set up Just
+        uses: extractions/setup-just@v1
+
       - name: Set up Docker Buildx
         id: buildx
         uses: docker/setup-buildx-action@v1
-
-      - name: Get version
-        id: version
-        run: |
-          # NOTE: --local ensures that "latest" builds use the version in our Makefile (instead of latest from the web)
-          VERSION=`./get-version.py ${{ matrix.config.product }} --type=${{ matrix.config.version }} --local`
-          SAFE_VERSION=`echo -n "$VERSION" | sed 's/+/-/g'`
-          echo "::set-output name=VERSION::${VERSION}"
-          echo "::set-output name=SAFE_VERSION::${SAFE_VERSION}"
-          if [[ "true" == "${{ github.ref == 'refs/heads/dev' }}" ]]; then
-            BRANCH_PREFIX="dev-"
-            echo "Setting BRANCH_PREFIX=$BRANCH_PREFIX-"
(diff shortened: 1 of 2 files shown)


Justfile
```diff
@@ -63,3 +65,99 @@ update-r-versions:
...
+
+# just get-version workbench --type=preview --local
+get-version +NARGS:
+  ./get-version.py {{NARGS}}
+
+# just get-safe-version workbench --type=preview --local
+get-safe-version +NARGS:
+  #!/usr/bin/env bash
+  VERSION=`./get-version.py {{NARGS}}`
+  echo -n "$VERSION" | sed 's/+/-/g'
```
atheriel, line 163: Nit: missing trailing newline.

.github/workflows/build-preview.yaml
```diff
@@ -13,60 +13,36 @@ on:
...
-    name: build-${{ matrix.config.tag_prefix }}${{ matrix.config.product }}:${{ matrix.config.version }}
-    needs: matrix
+    name: build-${{ matrix.config.product }}:${{ matrix.config.type }}:${{ matrix.config.os }}
 
     strategy:
       fail-fast: false
       matrix:
-        config: ${{ fromJson(needs.matrix.outputs.matrix) }}
+        config:
+         - {product: "workbench", type: "daily", os: "bionic"}
```
atheriel, line 25: FWIW, I'm a fan of hardcoding these. Simplicity is a win here.

atheriel (approved):
I don't feel totally comfortable assessing the new Bash code in the Justfile, but architecturally this makes sense to me.

.github/workflows/build-preview.yaml
```diff
@@ -75,48 +51,23 @@ jobs:
...
-          build-args: |
-            RSW_VERSION=${{ steps.version.outputs.VERSION }}
-            RSC_VERSION=${{ steps.version.outputs.VERSION }}
-            RSPM_VERSION=${{ steps.version.outputs.VERSION }}
-            # TODO: need a better way to handle pivoting DOWNLOAD_URL for OS
-            RSW_DOWNLOAD_URL=https://s3.amazonaws.com/rstudio-ide-build/server/${{ matrix.config.os }}/${{ matrix.config.arch }}
-
-      - name: Image digest
-        run: echo ${{ steps.docker_build.outputs.digest }}
+      - name: Build Image
```
colearendt, line 54: Are we doing anything to handle caching here? I believe the GHA had some niceties for that, but I admit to never having paid much attention

Justfile
```diff
@@ -63,3 +65,99 @@ update-r-versions:
...
+
+  # overwrite version if defined
+  if [[ $VERSION != "" ]]; then
+    REAL_VERSION=$VERSION
+    TAG_VERSION=`echo -n "$VERSION" | sed 's/+/-/g'`
+  fi
+
+  # set buildx args
+  if [[ "{{BUILDX_PATH}}" != "" ]]; then
+    BUILDX_ARGS="--cache-from=type=local,src=/tmp/.buildx-cache --cache-to=type=local,dest=/tmp/.buildx-cache"
```
colearendt, line 107: Is this sufficient for the caching, or do we need to add something into GHA so that it makes use of caching?

Justfile
```diff
@@ -63,3 +65,99 @@ update-r-versions:
...
+
+  docker buildx --builder="{{BUILDX_PATH}}" build --load $BUILDX_ARGS -t rstudio/rstudio-"$PRODUCT"-preview:"${BRANCH_PREFIX}${OS}"-"$TAG_VERSION" \
+        -t rstudio/rstudio-"$PRODUCT"-preview:"${BRANCH_PREFIX}${OS}"-"$TYPE" \
+        -t ghcr.io/rstudio/rstudio-"$PRODUCT"-preview:"${BRANCH_PREFIX}${OS}"-"$TAG_VERSION" \
+        -t ghcr.io/rstudio/rstudio-"$PRODUCT"-preview:"${BRANCH_PREFIX}${OS}"-"$TYPE" \
+        --build-arg "$SHORT_NAME"_VERSION=$REAL_VERSION \
+        --build-arg RSW_DOWNLOAD_URL=$RSW_DOWNLOAD_URL \
+        --file=./"$PRODUCT"/Dockerfile."$OS" "$PRODUCT"
+
+  echo rstudio/rstudio-"${PRODUCT}"-preview:"${BRANCH_PREFIX}${OS}"-"${TAG_VERSION}" \
```
colearendt, line 118: Maybe a comment to make clear that this is the _very important_ standard output to downstream processes?

Justfile
colearendt, line 163: Missing trailing newline

colearendt (commented):
Overall looking really good!

Justfile
dkasabovn: I looked at the actual docker/build plugin. This is what happens under the hood.

.github/workflows/build-preview.yaml
```diff
@@ -13,60 +13,36 @@ on:
...
         uses: actions/checkout@v2
 
+      - name: Set up Just
+        uses: extractions/setup-just@v1
+
       - name: Set up Docker Buildx
         id: buildx
         uses: docker/setup-buildx-action@v1
 
-      - name: Get version
```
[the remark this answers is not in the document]
dkasabovn: Not sure why I didn't think of this. Also removes the annoyance of having annoying logic in case you pass in a version. Will do this.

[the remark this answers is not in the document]
dkasabovn: By the above ^ I meant having a get-version step. Not sure about the second suggestion as for RSW a safe version is needed *alongside* the normal version. So I'm pretty sure just passing the real version down from a dedicated get-version step and creating a tag-safe version within a justfile would be ideal.

Justfile
ianpittwood: Do we need to use [actions/cache](https://github.com/marketplace/actions/cache) to make this work though?

dkasabovn: We have it, it is above the get-version step. I essentially copied the caching we previously had.

ianpittwood (approved):

Justfile
ianpittwood: Ah, I see it now. LGTM then!

colearendt (approved):
LGTM! Exciting stuff! 😄

PR #373, opened 2022-08-15
Refactor Build Latest

dkasabovn:
### Changes
Modify `build-preview-webhook.yaml` and only target single target builds.
Modify `build-latest.yaml`
Tidy up Justfile

### Motivation
build-latest: Refactored just like build-preview
build-preview-webhook: Refactored like build-preview and removed wildcard functionality. Now you can only perform one build at a time with the webhook.

### Testing
If you use [act](https://github.com/nektos/act) make sure you choose the largest image, otherwise docker-compose will not work. *this is a 20 gb image*

### Notes
`RSW_DOWNLOAD_URL` is now passed into every docker build


Justfile
```diff
@@ -66,14 +66,56 @@ update-r-versions:
   sed {{ sed_vars }} "s/^R_VERSION:.*/R_VERSION={{ R_VERSION }}/g" package-manager/Dockerfile.bionic
   sed {{ sed_vars }} "s|^RVersion.*=.*|RVersion = /opt/R/{{ R_VERSION }}/|g" package-manager/rstudio-pm.gcfg
 
+build-release $PRODUCT $OS $VERSION $BRANCH=`git branch --show` $SHA_SHORT=`git rev-parse --short HEAD`:
```
colearendt, line 69: Do you mind articulating how `build-release` differs from `build-preview`? It seems like there is a lot of duplicated logic there. I figure you evaluated the tradeoffs and why having a dedicated path makes sense? Is it just the `dev-branch` prefix mainly?

.github/workflows/build-preview-webhook.yaml
```diff
@@ -137,7 +88,4 @@ jobs:
       - name: Push image(s) to Docker Hub
         if: ${{ github.ref == 'refs/heads/main' || github.ref == 'refs/heads/dev' || github.ref == 'refs/heads/dev-rspm' }}
         run: |
-          docker push rstudio/${{ matrix.config.product }}:${{ steps.version.outputs.BRANCH_PREFIX }}${{ matrix.config.tag_prefix }}${{ matrix.config.version }}
-          docker push rstudio/${{ matrix.config.product }}:${{ steps.version.outputs.BRANCH_PREFIX }}${{ matrix.config.tag_prefix }}${{ steps.version.outputs.SAFE_VERSION }}
-          docker push ghcr.io/rstudio/${{ matrix.config.product }}:${{ steps.version.outputs.BRANCH_PREFIX }}${{ matrix.config.tag_prefix }}${{ matrix.config.version }}
-          docker push ghcr.io/rstudio/${{ matrix.config.product }}:${{ steps.version.outputs.BRANCH_PREFIX }}${{ matrix.config.tag_prefix }}${{ steps.version.outputs.SAFE_VERSION }}
+          just push-images ${{ steps.build-image.outputs.TAGS }}
```
colearendt, line 91: missing trailing newline

.github/workflows/build-preview-webhook.yaml
```diff
@@ -1,73 +1,43 @@
...
     inputs:
-      which:
-        description: "Which images to build. Default 'all'. Options: 'all','rstudio-workbench-preview','rstudio-connect-preview','rstudio-connect-content-init-preview','rstudio-package-manager-preview','r-session-complete-preview'. Can also use a prefix (i.e. 'rstudio-connect') to trigger multiple image builds."
+      product:
+        description: "Which image to build. Options: 'connect', 'connect-content-init', 'package-manager', 'r-session-complete', 'workbench', 'workbench-for-microsoft-azure-ml'."
         required: true
-        default: "all"
+        type: string
+      os:
+        description: "Which os to build. Default 'bionic'. Options: 'bionic', 'focal', 'centos7'"
```
colearendt, line 9: Should we leave off "focal" until we merge it? What is the behavior if you choose an image / OS combo that is invalid?

.github/workflows/build-latest.yaml
```diff
@@ -120,9 +85,4 @@ jobs:
       - name: Push image(s) to Docker Hub
         if: ${{ github.ref == 'refs/heads/main' }}
         run: |
-          docker push rstudio/${{ matrix.config.product }}:${{ matrix.config.tag_prefix }}${{ matrix.config.version }}
-          docker push rstudio/${{ matrix.config.product }}:${{ matrix.config.tag_prefix }}${{ steps.version.outputs.SAFE_VERSION }}
-          docker push rstudio/${{ matrix.config.product }}:${{ matrix.config.tag_prefix }}${{ steps.version.outputs.SAFE_VERSION }}${{ steps.version.outputs.SHA_SHORT }}
-          docker push ghcr.io/rstudio/${{ matrix.config.product }}:${{ matrix.config.tag_prefix }}${{ matrix.config.version }}
-          docker push ghcr.io/rstudio/${{ matrix.config.product }}:${{ matrix.config.tag_prefix }}${{ steps.version.outputs.SAFE_VERSION }}
-          docker push ghcr.io/rstudio/${{ matrix.config.product }}:${{ matrix.config.tag_prefix }}${{ steps.version.outputs.SAFE_VERSION }}${{ steps.version.outputs.SHA_SHORT }}
+          just push-images ${{ steps.build-image.outputs.TAGS }}
```
colearendt, line 88: Missing trailing newline

.github/workflows/build-latest.yaml
```diff
@@ -16,39 +16,25 @@ jobs:
...
-          - {product: 'r-session-complete', tag_prefix: 'bionic-', dir: 'r-session-complete', os: 'bionic', arch: 'amd64', prefix: RSW, version: 'latest'}
-          - {product: 'r-session-complete', tag_prefix: 'centos7-', dir: 'r-session-complete', os: 'centos7', arch: 'x86_64', prefix: RSW, version: 'latest'}
-          - {product: 'rstudio-workbench-for-microsoft-azure-ml', dir: 'workbench-for-microsoft-azure-ml', os: 'bionic', arch: 'amd64', prefix: RSW, version: 'latest'}
+          - {product: 'workbench', os: 'bionic'}
+          - {product: 'connect', os: 'bionic'}
+          - {product: 'connect-content-init', os: 'bionic'}
+          - {product: 'package-manager', os: 'bionic'}
+          - {product: 'r-session-complete', os: 'bionic'}
+          - {product: 'r-session-complete', os: 'centos7'}
+          - {product: 'workbench-for-microsoft-azure-ml', os: 'bionic'}
```
colearendt, line 25: This list is SOOOO much nicer!!!

Justfile
```diff
@@ -133,7 +174,7 @@ push-images +IMAGES:
   done
 
 # just test-image preview workbench 12.0.11-8 tag1 tag2 tag3 ...
-test-image $TYPE $PRODUCT $VERSION +IMAGES:
+test-image $PRODUCT $VERSION +IMAGES:
```
colearendt, line 177: I take it the "TYPE" arg wasn't used in testing?

colearendt (commented):
Looking really good!! Exciting stuff!

Justfile
dkasabovn: So there are three things.

1. Rsw build url
2. Tag formats
3. Branch prefixes

In addition since release does not need a type, it does not have the type parameter. To combine `build-preview` and `build-release` into one, the main obstacle would be generating tags based on build type and other variables. I think keeping the targets separate and hardcoding the tags is a bit clearer here.

.github/workflows/build-preview-webhook.yaml
dkasabovn: It will either fail at `get-version` or fail at `build` when the build target tries to stat the `Dockerfile`.

Justfile
dkasabovn: It was before when we generated version within test-image, but now we are passing version down, so we do not need it.

.github/workflows/build-preview-webhook.yaml
colearendt: Ahh nice, that makes sense. Which reminds me - we should add `os` to the "name" of the build below

.github/workflows/build-preview-webhook.yaml
```diff
@@ -1,73 +1,43 @@
...
     runs-on: ubuntu-latest
-    name: build-${{ matrix.config.tag_prefix }}${{ matrix.config.product }}:${{ matrix.config.version }}
-    needs: matrix
-    if: ${{ github.event.inputs.version == 'auto' || github.event.inputs.which != 'all' }}
-
-    strategy:
-      fail-fast: false
-      matrix:
-        config: ${{ fromJson(needs.matrix.outputs.matrix) }}
+    name: build-${{ github.event.inputs.product }}:${{ github.event.inputs.type }}-${{ github.event.inputs.version }}
```
colearendt, line 28: We should add OS here

Justfile
colearendt: Nice! Would you mind throwing this note into a `CONTRIBUTING.md` or `RELEASE.md` or `CI.md` or some such? I worry that we will forget this important info and it is easy to lose in PR feedback 😄

Justfile
```diff
@@ -66,14 +66,56 @@ update-r-versions:
...
+    SHORT_NAME="RSPM"
+  fi
+
+  # set buildx args
+  if [[ "{{BUILDX_PATH}}" != "" ]]; then
+    BUILDX_ARGS="--cache-from=type=local,src=/tmp/.buildx-cache --cache-to=type=local,dest=/tmp/.buildx-cache"
+  fi
+
+  docker buildx --builder="{{BUILDX_PATH}}" build --load $BUILDX_ARGS \
+        -t rstudio/rstudio-"$PRODUCT":"$OS"-latest \
```
colearendt, line 94: One question I naively never thought of until now. This will mean that we will no longer have a `latest` image so that `docker pull rstudio/rstudio-connect` will no longer do anything. Are we ok with this? We may want to call it out in `NEWS.md` and elsewhere. Are there conventions for this elsewhere?

colearendt: Also, I'm fine w/ doing docs PRs as a follow-up if need be. Just wanted to make sure the needs are captured 😄

dkasabovn: I think its ok to remove the generic image, especially since i'll be adding like 3 more OS' and people who are pulling the images should be cognizant of the choice they're making. But I can change as needed. And yes I think doing $OS-latest is redundant here, we should just do $OS.

Another option that we have is waiting for the new Jammy images I'll be pushing after this pr, and then setting the 'generic' image to the Jammy image as it'll probably be the most used. Lmk what you think.

colearendt: I like the idea of just `$OS` for now, and then we can figure out `latest` later.

Maybe worth adding a `NEWS.md` entry to the main `README.md` that CI is refactored, and that the `latest` image is not being built presently?

dkasabovn: Added

colearendt (approved):
LGTM! Let's get this puppy landed and iterate from there! 😄

PR #377, opened 2022-08-19
Fix makefile Dockerfile build targets

ianpittwood:
We recently changed our Dockerfile names to Dockerfile.bionic in preparation for additional OS support, but we did not update the Makefile and docker-compose Dockerfile targets to match this change. This PR updates all relevant places where the Dockerfile.bionic file must be specified. It also allows us to dynamically set the IMAGE_OS in the Makefile and main docker-compose.yml file. I also changed each build command to use `DOCKER_BUILDKIT=1` by default.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,7 +64,7 @@ update-drivers:  ## Update the driver version
 
 rsw: workbench
 workbench:  ## Build Workbench image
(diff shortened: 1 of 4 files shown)


atheriel (approved):
LGTM

Issue #117, opened 2021-05-10
Add a janitor / way to clean up images


colearendt:
There is a risk in leaving "old" images around too long. We should evaluate having a "janitor" to clean up old and defunct images

colearendt:
A downside here - it is nice having old versions around so that we can easily test old product versions when customers are running into issues in old environments 🤷‍♂️

ianpittwood:
I want to take this one on after I close out a couple other tickets. Docker Hub's [Advanced Image Management API](https://docs.docker.com/docker-hub/api/latest/#tag/images) seems like a decent way to do this, but we'll want to decide what criteria we want to prune on:

* Prune based off image age (i.e. image was pushed >2yrs ago then we can prune it).
* Prune based off last active time, the time since the image was last pushed or pulled. This is supported directly by the API with a default of 1 month.
* Some mix of the two.

We could also consider pruning any "preview" images at a faster clip than other images or even separating them into another feed (i.e. rstudio/workbench-preview) to make the distinction clearer for end users and so we can more easily apply different sets of rules to them. 

I'm leaning towards pruning based on last active time, but maybe at a further out timeline like 6 months to a year.

PR #379, opened 2022-08-19
Add an automation to cleanup old images/manifests/tags

ianpittwood:
Per #117, I implemented a script and Github Workflow to automatically delete images/manifests/tags that have been "inactive" (defined as no push or pull on the image) for the past year. The API endpoint used will not allow for "active" images to be deleted and we did not implement the `ignore_warnings` option that allows the call to bypass that safety feature. The Docker Hub API documentation can be found [here](https://docs.docker.com/docker-hub/api/latest/).

Currently the workflow has `DRY_RUN` hardcoded as enabled to prevent accidental changes prior to review. **This will need to be updated prior to merge.**


.github/workflows/clean-registry.yaml
```diff
@@ -0,0 +1,26 @@
+on:
+  schedule:
+    - cron: 0 0 * * 6
```
colearendt, line 3: Maybe add a comment explaining when this runs? Every Saturday at midnight UTC, right?

Is that when we want a potentially destructive operation running? 😅

.github/workflows/clean-registry.yaml
```diff
@@ -0,0 +1,26 @@
...
+  push:
```
colearendt, line 4: Why are we running on push and pull request too?

.github/workflows/clean-registry.yaml
```diff
@@ -0,0 +1,26 @@
...
+      - dev
+      - dev-rspm
+  pull_request:
+
+name: clean registries
+
+jobs:
+  cleanup:
+    env:
+      DOCKER_HUB_USERNAME: ${{ secrets.DOCKER_HUB_USERNAME }}
```
colearendt, line 16: Worth noting that this will not clean up ghcr.io, where we also push images, correct?

tools/dockerhub_clean.py
```diff
@@ -0,0 +1,145 @@
...
+            batch_count = 0
+            print("Deleting the following images:", file=sys.stderr)
+            pprint(print_image_list, stream=sys.stderr)
+            r = requests.post(ENDPOINTS["delete_images"], headers=headers, json=data)
+            if r.status_code != 200:
+                print(f"{r.status_code} Failed to delete batch of images for {repository}", file=sys.stderr)
+            data["manifests"] = []
+            print_image_list = []
+
+    if data["manifests"]:
```
colearendt, line 119: Maybe worth a comment to note that this is redundant on purpose? Am I right that it catches any "extras" that did not make it to the 25 count.

tools/dockerhub_clean.py
```diff
@@ -0,0 +1,145 @@
...
+        page += 1
+    return image_list
+
+
+def delete_images(bearer_token, repository, image_list, active_from, dry_run=True):
+    headers = {
+        "Authorization": f"Bearer {bearer_token}"
+    }
+    data = {
+        "dry_run": dry_run,
```
colearendt, line 70: I take it this is a post body argument we are sending to docker hub and relying on them to implement the dry run functionality?

tools/dockerhub_clean.py
```diff
@@ -0,0 +1,145 @@
...
+        "active_from": active_from,
+        "manifests": [],
+    }
+    print_image_list = []
+    total_images = 0
+    batch_count = 0
+    for image in image_list:
+        if image["status"] == "active":
```
colearendt, line 78: Is "active" a definition on DockerHub?

colearendt (commented):
Very nice!! My main concern is around the dependencies on docker hub's implementation (i.e. dry run and "active" definition, etc.). This is a scary thing to go sideways (and clear out your repository / take out all customer implementations depending on these images, including our own). Moreover, while "not pulled in a year" seems like it would be sufficient, it also means that it is possible that docker images for versions that we _currently support_ could also be reaped (support lasts 18 months). Is this concerning to folks? Especially since rebuilding those images would likely be pretty tricky given our current architecture?

We also have the same problem on ghcr.io. I wonder whether ghcr plans to implement their own lifecycle policies (like ECR does) at some point.

.github/workflows/clean-registry.yaml
ianpittwood: You can treat it as a placeholder. By default I schedule things on the weekends, but I'm open to do it any other time(s) as well.

.github/workflows/clean-registry.yaml
ianpittwood: Another placeholder. I was between PR and cron, but I mostly added this so I was triggering the dry run every time I pushed. My Docker Hub credentials don't have high enough permissions to do any testing locally.

.github/workflows/clean-registry.yaml
ianpittwood: Yes, the thought behind that was that we can remove older images we no longer wish for people to start out using from Docker Hub where they're more publicly advertised, but keep a longer history of images in GHCR in case we need to go back and troubleshoot something on an older image for an end user. I'm assuming we could implement the same behavior on GHCR though if we want to.

tools/dockerhub_clean.py
ianpittwood: Yup, I'll add that.

tools/dockerhub_clean.py
ianpittwood: Yes, it's part of their API spec to check for errors and warnings on the calls without deleting any images: https://docs.docker.com/docker-hub/api/latest/#tag/images/operation/PostNamespacesDeleteImages

tools/dockerhub_clean.py
ianpittwood: Yes, Docker's default definition is that the image "has been pushed or pulled within the last month", but the time can be adjusted by in most of the API calls with the `active_from` parameter. I changed it to a year for the rough draft of the script.

ianpittwood:
> Moreover, while "not pulled in a year" seems like it would be sufficient, it also means that it is possible that docker images for versions that we _currently support_ could also be reaped (support lasts 18 months). Is this concerning to folks? Especially since rebuilding those old images would likely be pretty tricky given our current architecture?

This is helpful info! I'll change the default for `DAYS_SINCE_LAST_ACTIVE` (which overrides Docker's 30 day definition) from 365 days to 548 days. That should prevent anything within support from being reaped.

I also have a couple extra safeguards embedded in the `delete_images` function to double check that the image is not active (`get_images` filters on this already and Docker issues a warning if you attempt to delete an active image too) and that the image is not currently tagged as `latest`. If there are other safeguards we should implement, I'd be happy to do so.

> We also have the same problem on ghcr.io. I wonder whether ghcr plans to implement their own lifecycle policies (like ECR does) at some point.

There are some Github Actions out there to perform similar tasks: https://github.com/snok/container-retention-policy

Maybe we could look into implementing something like that?

.github/workflows/clean-registry.yaml
colearendt: Yeah I'm inclined to schedule this early in the week... maybe not Monday? So Tuesday? Haha. Lots of holidays on Mondays. I'd just hate for something to go awry and pull someone into working on Saturday / Sunday to figure out what went wrong. Weekends seem great for non-destructive operations that take build time out of a finite queue, but for something destructive, working hours seems best 😄

.github/workflows/clean-registry.yaml
colearendt: Yeah, I'd be inclined to remove this stuff before merging 😄

.github/workflows/clean-registry.yaml
```diff
@@ -0,0 +1,26 @@
...
+      DOCKER_HUB_PASSWORD: ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
+      DRY_RUN: 1
```
colearendt, line 18: I think merging this with DRY_RUN on is a good call. Then we can pivot to DRY_RUN: 0 when we're ready to lay waste to the old images!

colearendt (commented):
Just a few nits on the cron stuff 😄

ianpittwood:
> Just a few nits on the cron stuff smile

Should be changed to only run on Tuesdays at midnight now!

.github/workflows/clean-registry.yaml
ianpittwood: Should we perhaps change this to a secret similar to the two above? Something like `DOCKER_HUB_CLEAN_DRY_RUN` that way it's easier for us to flip on and off?

colearendt: My thought was to just merge as-is and then flip it on once verified that things are working as expected. A secret feels weird since it is really just a "repo-wide variable." And "repo-wide variable" has the same scope as code 😄

colearendt (approved):
LGTM!

PR #380, opened 2022-08-22
add quarto for python 3.9

shahmonanshi:


content/base/bionic/maybe_install_quarto.sh
```diff
@@ -1,8 +1,8 @@
 #!/bin/bash
 
-# only install quarto if python 3.10 and R 4.1
+# only install quarto if python 3.10 and R 4.1 OR python 3.9 and R 4.1
 # TODO: figure out a different hierarchy...
-if [[ `ls /opt/python/ | grep '3\.10\.'` ]] && [[ `ls /opt/R | grep '4\.1\.'` ]]; then
+if [[ `ls /opt/python/ | grep '3\.10\.'` ]] && [[ `ls /opt/R | grep '4\.1\.'` ]] || [[ `ls /opt/python/ | grep '3\.9\.'` ]] && [[ `ls /opt/R | grep '4\.1\.'` ]] ; then
```
colearendt, line 5: Have you tested the order of operations / precedence here for `&&` vs. `||` to make sure all of the peers work as expected?

colearendt (commented):
@aronatkins thoughts on adding another version of quarto for python 3.9? This was requested internally for some demo assets.

aronatkins (commented):
No concern with adding Quarto to additional containers. Maybe make this explicit, somehow?

content/base/bionic/maybe_install_quarto.sh
shahmonanshi: just added in a second set of brackets to absolutely ensure it does what we want it to do :)

bschwedler:
Quarto is always installed in Ubuntu 20.04 now.
https://github.com/rstudio/rstudio-docker-products/blob/ded71ad076e48cab9ae2a5e8a0af1d6714c1e16b/content/base/maybe_install_quarto.sh#L13-L21

PR #383, opened 2022-08-25
Bump Connect to 2022.08.1

marcosnav:
Pushed along a small adjustment to the `Justfile`. The `update-versions` command was failing due to an incomplete Dockerfile name.


colearendt (approved):
LGTM! Thanks!

PR #384, opened 2022-08-26
Implement linting on dockerfiles

ianpittwood:
To improve our code quality and adherence to Docker best practices, this pull request implements [Hadolint](https://github.com/hadolint/hadolint) in our workflow. The minimum failure threshold is set to `warning` in the Github Workflow so `style` and `info` suggestions will not fail a PR, though I encourage future contributors to consider using them!

I did my best to implement all the changes suggested by Hadolint, but I did exclude or ignore the following rules from implementation:
- Ignored
  - DL3008 - `warning` for unpinned packages in `apt-get install` commands. We don't currently practice dependency pinning and it would be a large undertaking to do so.
  - DL3033 - `warning` for unpinned packages in `yum install` commands.
  - DL3059 - `info` for multiple consecutive RUN statements. Many of these are conscious design choices, but they may still clutter PRs with comments. For now we'll officially ignore them, but we could look into condensing them down the road.
- Excluded (not explicitly ignored, but I did not implement them)
  - DL3059 - `info` for multiple consecutive `RUN` instructions. We intentionally use multiple `RUN` instructions in most Dockerfiles so it would not make sense to consolidate these at this time.

### DIFF
--- a/connect-content-init/Makefile
+++ b/connect-content-init/Makefile
@@ -8,17 +8,21 @@
 #     make RSC_VERSION=<version string>
 
 TAG_BASE?=rstudio/rstudio-connect-content-init-preview
-RSC_VERSION?=1.8.8.3-dev312
+RSC_VERSION?=2022.08.1
 
 all: build
 .PHONY: all
 
 # Builds the image using a given RSC distribution for runtime artifacts.
 build:
-	docker build --build-arg RSC_VERSION=$(RSC_VERSION) -t $(TAG_BASE):$(RSC_VERSION) .
+	DOCKER_BUILDKIT=1 docker build --build-arg RSC_VERSION=$(RSC_VERSION) -t $(TAG_BASE):$(RSC_VERSION) -f Dockerfile.bionic .
 .PHONY: build
 
 # Do not run push if you are not Cole.
 push: build
 	docker push $(TAG_BASE):$(RSC_VERSION)
 .PHONY: push
+
+test:
+	IMAGE_NAME=$(TAG_BASE):$(RSC_VERSION) docker-compose -f docker-compose.test.yml run sut
+.PHONY: test
--- a/hadolint.yaml
+++ b/hadolint.yaml
@@ -1,0 +1,6 @@
+failure-threshold: warning
+ignored:
+  - DL3008  # Suppresses warning for version pinning on apt installs
+  - DL3033  # Suppresses warning for version pinning on yum installs
+label-schema:
+  maintainer: text
--- a/hadolint.yaml
+++ b/hadolint.yaml
@@ -1,6 +1,11 @@
 failure-threshold: warning
 ignored:
-  - DL3008  # Suppresses warning for version pinning on apt installs
-  - DL3033  # Suppresses warning for version pinning on yum installs
+    # Suppresses warning for version pinning on apt installs. We currently do not practice this approac
+  - DL3008
+    # Suppresses warning for version pinning on yum installs. We currently do not practice this approach.
+  - DL3033
+    # Suppresses info for too many consecutive RUN statements. Multiple consecutive RUN statements are often a conscious
+    # design choice in our Dockerfiles, but we could consider condensing them in the future.
+  - DL3059
 label-schema:
   maintainer: text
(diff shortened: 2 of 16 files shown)


cm421:
Very minor nit: Any reasons _not_ to ignore DL3059? Even though they're info-level and below the failure threshold, they still appear when looking at the files changed in the PR (which is really cool they pop up like that!). This PR alone has 4, which might drown out legitimate errors in the future if I just get used to ignoring them.

cm421 (approved):
Very minor nit with DL3059 but otherwise LGTM!

ianpittwood:
> Very minor nit: Any reasons _not_ to ignore DL3059? Even though they're info-level and below the failure threshold, they still appear when looking at the files changed in the PR (which is really cool they pop up like that!). This PR alone has 4, which might drown out legitimate errors in the future if I just get used to ignoring them.
> 
> 
> 
> Also, [Hadolint's own dos](https://github.com/hadolint/hadolint/wiki/DL3059) suggests that check is a little extra:
> 
> 
> 
> > You might want to ignore this info level rule.

Yeah I only avoided ignoring it since we may want to implement it in the future, but since it does appear in the comments (also agree that's awesome, I didn't even set that up) I think you're right that we should silence it.

hadolint.yaml
```diff
@@ -1,6 +1,11 @@
 failure-threshold: warning
 ignored:
-  - DL3008  # Suppresses warning for version pinning on apt installs
-  - DL3033  # Suppresses warning for version pinning on yum installs
+    # Suppresses warning for version pinning on apt installs. We currently do not practice this approac
```
colearendt, line 3:
```suggestion
    # Suppresses warning for version pinning on apt installs. We currently do not practice this approach.
```

colearendt (approved):
Total nit / spelling mistake 😄

connect-content-init/Makefile
```diff
@@ -8,17 +8,21 @@
 #     make RSC_VERSION=<version string>
 
 TAG_BASE?=rstudio/rstudio-connect-content-init-preview
-RSC_VERSION?=1.8.8.3-dev312
+RSC_VERSION?=2022.08.1
```
colearendt, line 11: We should probably make sure this is caught in our regex

hadolint.yaml
```diff
@@ -0,0 +1,11 @@
+failure-threshold: warning
+ignored:
+    # Suppresses warning for version pinning on apt installs. We currently do not practice this approac
```
colearendt, line 3: Same

r-session-complete/Dockerfile.bionic
```diff
@@ -7,11 +7,13 @@ ARG R_VERSION=4.1.0
 ARG MINICONDA_VERSION=py37_4.8.3
 ARG PYTHON_VERSION=3.9.5
 ARG DRIVERS_VERSION=2021.10.0
+ENV DEBIAN_FRONTEND=noninteractive
```
colearendt, line 10: Shouldn't this be an ARG? I think ENV leaves it in the container forever?

colearendt (commented):
Sorry - a few more thoughts. I didn't see all the other commits 🙈

connect-content-init/Makefile
ianpittwood: I actually did not mean to commit this, but I can make that update. I wasn't sure if this was intentionally excluded tbh.

r-session-complete/Dockerfile.bionic
ianpittwood: Good point. I looked it up and it's heavily suggested not to leave this as an artifact in your containers. I added `ENV DEBIAN_FRONTEND=dialog` to the end of all the Dockerfile.bionic definitions to reset it back to default for future users of the container.

shepherdjerred (approved):

shepherdjerred:
I haven't seen hadolint before; it looks great. I'm a big fan of linting, so this gets a big thumbs up from me!

colearendt (approved):
LGTM!

melissa-barca (approved):
LGTM!

PR #385, opened 2022-08-29
Add centos7

dkasabovn:
# Changes
- Remove hardcoded gid and uid from GOSS yaml file due to os-specific user allocation
- Add centos7 images to all release builds

# Details
A flag is passed to `yum install` operations to ensure that we do not have ignored packages listed in the `Dockerfile`. If a package is not found it will throw an error.


colearendt:
I probably should have thought about this more deeply when we talked about it before 🙈 Are we sure we want to add centos7 images? They are EOL in 2024, and it will be ultimately setting users up for needing to switch their OS (which is a pretty intensive operation).

dkasabovn:
> I probably should have thought about this more deeply when we talked about it before see_no_evil Are we sure we want to add centos7 images? They are EOL in 2024, and it will be ultimately setting users up for needing to switch their OS (which is a pretty intensive operation).

No worries will just rebase the rl8 branch and close this one

PR #390, opened 2022-09-09
Trying to fix Issue with PAM chpasswd and SELinux

SOLDIERz:
This should solve the Issue which I was running into with OpenShift/Container Environments in the combination with PAM and chpasswd. I already opened an Issue for that in the Helm Charts Repo for RStudio: https://github.com/rstudio/helm/issues/265


colearendt:
Thanks for submitting this @SOLDIERz ! Have you tested this in an SELinux environment and found that it worked well?

PR #396, opened 2022-09-14
change session installation order

colearendt:
Installing Workbench last, so that R and Python versions are the "more immutable" parts of the image


cm421 (approved):
LGTM!

colearendt:
One thing we are considering to add here is building FROM our newer `content-base` or `content-pro` images, which will require a bit of change in how we structure our CI, so that these session images can build _after_ those images.

ianpittwood (approved):
Makes sense to me!

colearendt:
I'm thinking we may use #415 in lieu of this PR - it decouples in the same direction, but into completely different files / docker images.

colearendt:
Closed in favor of #454 (which includes #415)

PR #402, opened 2022-09-29
Refactor all build and test actions into Make

ianpittwood:


.github/workflows/build-latest.yaml
```diff
@@ -52,7 +52,7 @@ jobs:
       - name: Build Image
         id: build-image
         run: |
-          TAGS=`just BUILDX_PATH=${{ steps.buildx.outputs.name }} build-release ${{ matrix.config.product }} ${{ matrix.config.os }} ${{ steps.get-version.outputs.VERSION }}`
```
colearendt, line 55: So y'all are dropping `just` completely?

Justfile
```diff
@@ -57,127 +71,72 @@ update-rsc-versions:
...
   set -euxo pipefail
+  sed {{ sed_vars }} "s/^DRIVERS_VERSION=.*/DRIVERS_VERSION={{ DRIVERS_VERSION }}/g" content/pro/Makefile
+  sed {{ sed_vars }} "s/\"drivers\": \".[^\,\}]*\"/\"drivers\": \"{{ DRIVERS_VERSION }}\"/g" content/matrix.json
+  sed {{ sed_vars }} "s/^ARG DRIVERS_VERSION=.*/ARG DRIVERS_VERSION={{ DRIVERS_VERSION }}/g" helper/workbench-for-microsoft-azure-ml/Dockerfile
+  sed {{ sed_vars }} "s/^DRIVERS_VERSION ?= .*/DRIVERS_VERSION ?= {{ DRIVERS_VERSION }}/g" Makefile
+  sed {{ sed_vars }} "s/^DRIVERS_VERSION := .*/DRIVERS_VERSION := \"{{ DRIVERS_VERSION }}\"/g" Justfile
+
+build-release $PRODUCT $OS:
+  #!/usr/bin/env bash
+  make PRODUCT=${PRODUCT} IMAGE_OS=${OS} build
```
colearendt, line 134: just calling make

Makefile
```diff
@@ -47,59 +51,206 @@ else ifeq ($(UNAME_S),Darwin)
...
+	$(MAKE) PRODUCT=connect IMAGE_OS=$(IMAGE_OS) test
+	$(MAKE) PRODUCT=package-manager IMAGE_OS=$(IMAGE_OS) test
+	$(MAKE) PRODUCT=r-session-complete IMAGE_OS=$(IMAGE_OS) test
+	$(MAKE) PRODUCT=workbench IMAGE_OS=$(IMAGE_OS) test
+	$(MAKE) PRODUCT=workbench-for-microsoft-azure-ml IMAGE_OS=$(IMAGE_OS) test
+
+
+### Update versions shortcuts ###
+update-versions:  ## Update the versions for all products
+	just \
```
colearendt, line 241: make calling just

colearendt (commented):
Honestly, I feel like this approach makes things more confusing and harder to follow:

- `Justfile` works well for complicated situations because it can have `python` or other scripts in addition to `bash` scripts, quoting that doesn't read like a bashism, etc. However, it is not as portable to customers
- `Makefile` is ubiquitous, so more useful in customer environments

- Our CI is complicated because we build lots of variations of tags, versions, etc. Customers rarely / never need this complexity. We should hide it from them
- we don't have any clear data that customers even use the `Makefile`. Its main value is as a list of "commands you can easily run." Most folks probably cd into the container directory and `docker build`

As such, along with the fact that we have weird circular dependencies (`make` call `just`, `just` call `make`, which is the parent, etc.), complexity in the Makefile, etc. I think we should:

- ensure `just` and `make` handle different scopes of work
- use `just` for CI and complex / developer machinery
- use `make` for dead-simple examples of how to run things or build very simple cases. The target user is a novice docker user who might read the `Makefile` to see how we run commands. Abstractions and complexity will lose them
- minimize `make` calling `just` or `just` calling `make`. Again, if their audiences are different, there is no reason for either one to be "the complete answer." I don't think we should target "both just and make users"

.github/workflows/build-latest.yaml
ianpittwood: Nope, we still use `just` in other spots. I could alias this one too if it's a problem.

ianpittwood:
> Honestly, I feel like this approach makes things more confusing and harder to follow:
> 
>     * `Justfile` works well for complicated situations because it can have `python` or other scripts in addition to `bash` scripts, quoting that doesn't read like a bashism, etc. However, it is not as portable to customers
> 
>     * `Makefile` is ubiquitous, so more useful in customer environments

Agreed, but personally I'd always lean towards making things more accessible to end users.

>     * Our CI is complicated because we build lots of variations of tags, versions, etc. Customers rarely / never need this complexity. We should hide it from them

We don't currently hide any of this in `just` either so we'd likely need to alter our approach altogether if we want to do that. Maybe by feeding tags into make/just at build time in GHA or from another target?

>     * we don't have any clear data that customers even use the `Makefile`. Its main value is as a list of "commands you can easily run." Most folks probably cd into the container directory and `docker build`

We reference the `make` targets throughout the docs still for anyone following that. We can change that to `just`, but we'll want to include enough info for anyone who can't use `just` in their environment to be able to perform basic build/run commands. `just` is fairly straight forward, but links to documentation should help ensure we're not leaving anyone OOTL.

> As such, along with the fact that we have weird circular dependencies (`make` call `just`, `just` call `make`, which is the parent, etc.), complexity in the Makefile, etc. I think we should:
> 
>     * ensure `just` and `make` handle different scopes of work

That's my intention. This is still WIP, but my goal is to avoid circular dependencies. Calls from `make` to `just` or vice versa are really just alias wrappers so users can choose either tool. I can completely remove the aliases if it's a problem.

>     * use `just` for CI and complex / developer machinery
> 
>     * use `make` for dead-simple examples of how to run things or build very simple cases. The target user is a novice docker user who might read the `Makefile` to see how we run commands. Abstractions and complexity will lose them
> 
>     * minimize `make` calling `just` or `just` calling `make`. Again, if their audiences are different, there is no reason for either one to be "the complete answer." I don't think we should target "both just and make users"
> 
> 
> There is another path where we discard `just` in lieu of `make`. I am personally not a fan of that idea, because `make` is a bit of a pain for this type of command-line interface IMO. Further, then we get complexity in the `Makefile` which makes it inscrutable to customers. I'm going to ask the RSC team for a bit of a retrospective on their thoughts after abandoning `make`.

I'm fine with fully abandoning one tool for the other as it will clean things up significantly. Personally though, I think we should keep `make` since it's the universal solution to this problem. That being said, I can redo this for whatever our tool of choice ends up being.

We could also consider breaking up the make/just files into the sub-directories to make things a little more readable.

ianpittwood:
Starting new PR for full switch to Just.

PR #404, opened 2022-10-04
Adjust Snyk to test on branch and keep monitor on main

zfouts:
Move snyk scans to "test" for branches/PR's and keep "monitor" on main branch


colearendt (approved):
LGTM! Just to understand, the idea here is that there will not be notifications from branches?

PR #412, opened 2022-10-20
fix a typo in the content-pro dockerfile build

colearendt:
We accidentally created a recursive reference from content-pro images to other content-pro images. The reference here should have been content-base.


aronatkins (approved):
Recursion is fun!

Issue #371, opened 2022-08-11
Review pro product image tagging strategy


ianpittwood:
I've noticed that we're lacking a lot of "high level" tagging for our pro product images. One of the niceties of Docker is that we can present "simple" tags at the forefront of our images to customers while still allowing them to dig down through more specific tags, all the way down to the digest.

For example, here's a way we could add higher level tags to `rstudio/rstudio-workbench`:
* `latest`
* `2022.07.1` (w/ multi-os support we could format this as `2022.07.1-bionic` with a default image set for `2022.07.1`)
* `2022.07.1-554.pro3`
* `2022.07.1-554.pro3--6424ea7` or push users that need these hyper-specific tags towards using the SHA256 digest reference instead

This is likely something we can overhaul from `r-base` all the way up to our pro product images.

We should also consider how we can modify building and image tagging to allow for users to easily select from different Python and R version combinations. (related to #222)

ianpittwood:
@kfeinauer @colearendt thoughts?

bdeitte:
Whoever takes on this issue will review some internal documents on this topic, then we'll see if there's anything more we would like to do here for better labeling.  After that, we can make any changes (if needed) and document what is there (if needed).

colearendt:
We chatted about this offline IIRC - we should:
- strip the build metadata from the image tag in Workbench & RSPM (like we do for the  `azure` image)
- Keep the `--SHA` variant for more immutable images
- add labels for R versions, python versions, etc.

I.e. to use your workbench example, I think we only need:
- `bionic`
- `bionic-2022.07.1`
- `bionic-2022.07.1--6424ea7`

Having a "default" `latest`, `2022.07.1`, etc. is discussed in #381 and #399

PR #416, opened 2022-10-24
Simplify Docker Tagging

zfouts:
This PR addresses https://github.com/rstudio/rstudio-docker-products/issues/371

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,10 @@ changed in each image.
 
 This file only captures pervasive, repository-wide changes.
 
+# 2022-11-07
+- Potentially breaking changes around tagging with workbench due to simplifying the tagging approach.
+- Remove internal version strings e.g. `554.pro3`.
+
 # 2022-10-12
 - We replaced all `make` definitions with [`just`](https://just.systems/man/en) for ease of use.
 


colearendt (changes requested):
Unfortunately, I don't think this is going to have the desired effect. We used to have this taken care of for the Azure image, but it seems to have been dropped during the CI refactor. https://github.com/orgs/rstudio/packages/container/rstudio-workbench-for-microsoft-azure-ml/versions

NOTE: _old_ commit reference
https://github.com/rstudio/rstudio-docker-products/blob/4fd848eb0baac0bd56d28a6dc7d32796b03cae26/.github/workflows/build-latest.yaml#L43-L45

It looks like the `RSW_TAG_SAFE_VERSION` is only used in the compose file here. You may need to take a look at the `ci.Justfile`? Or maybe the child `Justfiles` now too?

https://github.com/rstudio/rstudio-docker-products/blob/5fdd3c0a09b8494e6980ef31619b464149d9a956/ci.Justfile#L129-L131

It's also worth noting that while this will resolve things for RSW, it will not solve the problem for RSPM, which has a `-` that we would ultimately want to drop: 
https://github.com/rstudio/rstudio-docker-products/pkgs/container/rstudio-package-manager

RSC does this work for us, so isn't a great example (i.e. in your comment on the issue). RSW and RSPM are the troublemakers that include their build metadata in their released version number 😉

Justfile
```diff
@@ -9,7 +9,7 @@ BUILDX_PATH := ""
 RSC_VERSION := "2022.09.0"
 RSPM_VERSION := "2022.07.2-11"
 RSW_VERSION := "2022.07.2+576.pro12"
-RSC_TAG_SAFE_VERSION := replace(RSW_VERSION, "+", "-")
+RSW_TAG_SAFE_VERSION := replace(RSW_VERSION, "+.*$", "")
```
ianpittwood, line 12: [`replace` and other similar functions in `just`](https://just.systems/man/en/chapter_30.html#string-manipulation) don't yet support regex despite having the misleading argument name of `pattern`. I tried messing with something like 

    RSW_TAG_SAFE_VERSION := `echo {{RSW_VERSION}} | sed 's/+.*$//g'`

, but it seems `just` doesn't support using a variable inside a command eval like that.

Justfile
```diff
@@ -46,7 +46,7 @@ update-rsw-versions:
     workbench/Dockerfile.bionic \
     workbench-for-microsoft-azure-ml/Dockerfile.bionic
   sed {{ sed_vars }} "s/RSW_VERSION:.*/RSW_VERSION: {{ RSW_VERSION }}/g" docker-compose.yml
-  sed {{ sed_vars }} "s/rstudio\/rstudio-workbench:.*/rstudio\/rstudio-workbench:{{ RSC_TAG_SAFE_VERSION }}/g" docker-compose.yml
+  sed {{ sed_vars }} "s/rstudio\/rstudio-workbench:.*/rstudio\/rstudio-workbench:{{ RSW_TAG_SAFE_VERSION }}/g" docker-compose.yml
```
ianpittwood, line 49: Good catch!

ianpittwood (changes requested):
So we'll want to make these changes in a few more places:
* ci.Justfile:139
* workbench-for-microsoft-azure-ml/Justfile:10
* workbench/Justfile:10

Unfortunately we'll need to use a `just` target to do this I think since command eval doesn't work with variables. Maybe we can update and move `_get-tag-safe-version` from ci.Justfile to Justfile and have everything reference that target? For the subdirectory Justfiles we can use the [`justfile_directory()` function](https://just.systems/man/en/chapter_30.html#justfile-and-justfile-directory) to target the main Justfile relative to that directory I think.

Justfile
```diff
@@ -20,6 +19,11 @@ R_VERSION_ALT := "4.1.0"
 PYTHON_VERSION := "3.9.5"
 PYTHON_VERSION_ALT := "3.8.10"
 
+# just _get-tag-safe-version 2022.07.2+576.pro12
+_get-tag-safe-version $VERSION:
```
kfeinauer, line 23: Here and elsewhere - underscore in the just target seems a bit odd.

kfeinauer: Are we going to do something similar for other product (RSC/RSPM) versions to strip off build number?

AlexMapley: I'd prefer `get-tag-safe-version` as well.

Justfile
```diff
@@ -46,7 +50,7 @@ update-rsw-versions:
     workbench/Dockerfile.bionic \
     workbench-for-microsoft-azure-ml/Dockerfile.bionic
   sed {{ sed_vars }} "s/RSW_VERSION:.*/RSW_VERSION: {{ RSW_VERSION }}/g" docker-compose.yml
-  sed {{ sed_vars }} "s/rstudio\/rstudio-workbench:.*/rstudio\/rstudio-workbench:{{ RSC_TAG_SAFE_VERSION }}/g" docker-compose.yml
+  sed {{ sed_vars }} "s/rstudio\/rstudio-workbench:.*/rstudio\/rstudio-workbench:$(just _get-tag-safe-version {{ RSW_VERSION }})/g" docker-compose.yml
```
AlexMapley, line 53: I'm not too familiar with best `just`  practices, do we worry about these recursive calls where the subcommands of the `Justfile` are executing different just subcommands?

I think this patterns works well for things like `update-versions` which wraps a call to `just update-rsw-versions`, but it does make the justfile more tangly in general whenever it's calling itself.

The `get-tag-safe-version` command seems helpful elsewhere, but I'd support readding that `RSC_TAG_SAFE_VERSION` var instead even if we have to duplicate a couple lines of code.

Justfile
ianpittwood: > Here and elsewhere - underscore in the just target seems a bit odd.

I preface the targets that are for internal use with an `_`, similar to how it's used in Python. People typically won't use this one directly, it's just reused by other targets.

> Are we going to do something similar for other product (RSC/RSPM) versions to strip off build number?

RSC already strips off everything to give us a pure `YYYY.MM.DD` version. RSPM adds a `-ZZ` build number onto that which we could strip off if desired. RSW was just the main concern because everything after the `+` is confusing metadata if someone just needs a specific version.

Justfile
ianpittwood: OK so I agree, it's weird. But the only reason I went this route is because we can't make variables with command evals that use other variables in Just. I opened an issue with Just (see casey/just#1388) to see about adding that behavior. We also can't use one of the more straightforward functions like `replace` or `trim_end_match` because they don't support regex, which I also opened an issue for (see casey/just#1389).

ianpittwood: > The `get-tag-safe-version` command seems helpful elsewhere, but I'd support readding that `RSC_TAG_SAFE_VERSION` var instead even if we have to duplicate a couple lines of code.

Also important to note that we can't accomplish fully stripping off the `+` metadata by going back to `RSW_TAG_SAFE_VERSION`. We would have to put the version number in both `RSW_VERSION` (raw) and `RSW_TAG_SAFE_VERSION` (wrapped in a sed command) to get the expected results and update our regex find/replace to match those updates. I thought using a new just target would be the least complex way to implement the change.

Justfile
kfeinauer: I would say go ahead with stripping off RSPM build number to make everything consistent and this PR LGTM!

kfeinauer (approved):

Justfile
AlexMapley: If we're acknowledging the tradeoffs, I'm good to go 👍. Ty for your detailed explanation.

I almost didn't make this comment too, I could still understand what was happening in the code. If we have to get a little creative because we're brushing up against the limitations of the framework, that's totally acceptable.

AlexMapley (approved):
lgtm. It looks like those 5 ci tests we're failing are also failing on main, so doesn't seem like a blocker for this PR.
(we might want to prioritize fixing those though)

ianpittwood:
> lgtm. It looks like those 5 ci tests we're failing are also failing on main, so doesn't seem like a blocker for this PR. (we might want to prioritize fixing those though)

It's fixed in `dev`. `workbench` made an update that preinstalls VS Code, but it also causes all our previews/dailies to fail when they try to install VS Code. We'll merge the fix to remove the VS Code install commands on their next release.

Justfile
ianpittwood: @kfeinauer should be good to go now.

kfeinauer: LGTM!

colearendt (approved):
LGTM! So exciting to have this done! 

The only caveat I would say - please add `NEWS.md` entries on the images who changed and one on the main `NEWS.md` file, so we have somewhere to point if/when folks wonder why they are not getting the most updated image anymore (i.e. existing tags become "orphans" in some sense - no more updates will come to it). 

We will need to update the helm charts to target these image tags too. 

We should also probably ensure that the Workbench and RSPM teams are aware that versioned tags will look a smidge differently now - an offline ping / PSA should suffice IMO 😄 I don't think it affects any of their automations AFAIK, which often use `daily` type of tags... but there may be some places they reference an explicit version.

tylfin:
@zfouts coming in here late because we were doing a release (e.g. pinned to a specific docker image). We're not seeing daily build images, e.g. [bionic-2022.07.3-1187](https://hub.docker.com/layers/rstudio/rstudio-package-manager-preview/bionic-2022.07.3-1187/images/sha256-4d529cfe861047637e5a806e5999617ef16620b13380e8db90af96c9cb9154f9?context=explore) vs. [bionic-2022.11.1](https://hub.docker.com/layers/rstudio/rstudio-package-manager-preview/bionic-2022.11.1/images/sha256-197b179e7bda46c7ab66f5263b54592b1283af4004467215fb6e6d3c8cce0aec?context=explore). This is impacting our K8s QA sites as it won't automatically repull these images without a force kill. Did this PR discontinue those builds?

tylfin:
Stripping the build number is fine, but I've typically seen both tagging strategies co-existing, e.g. https://hub.docker.com/_/ubuntu/tags

PR #420, opened 2022-10-27
Change ubuntu codenames to full version numbers

ianpittwood:
This change swaps the majority of Dockerfiles over to full version number extensions instead of code name extensions for Ubuntu (i.e. `Dockerfile.ubuntu1804` from `Dockerfile.bionic`). I think using the full name and version is preferable for the following reasons:
* A code name, such as `bionic`, does not convey what the underlying OS is at a glance for end users. `ubuntu1804` conveys both the OS and the version of the OS.
* Users who are not heavy Linux/Ubuntu users may not have the Ubuntu code names committed to memory.
* We only use code names to refer to Ubuntu, no other OSes. For instance, we don't refer to `centos7` as just `7`. `OSVersionNumber` is more consistent.

Despite changing the Ubuntu Dockerfile extensions and primary references to `ubuntuYYMM`, this change retains the codename tags alongside the new `ubuntuYYMM` tags.

The changes were not applied to the `content` images for now as they are behind on other updates from a previous refactor. The changes will be added to those images in a later change.


colearendt (commented):
Generally looks good. How did you decide between `ubuntu18`, `ubuntu1804`, and `ubuntu18.04`?

colearendt (approved):
LGTM! I'd say change immediately after merging 😄  Because it will immediately affect other PRs / etc.

Apologies for never responding before on the naming - that makes sense. It seems like dots are somewhat conventional in the places I checked ([here](https://hub.docker.com/_/redis) and [here](https://hub.docker.com/_/postgres)), but I kinda like the dotless version 😄

ianpittwood:
> LGTM! I'd say change immediately after merging smile Because it will immediately affect other PRs / etc.
> 
> Apologies for never responding before on the naming - that makes sense. It seems like dots are somewhat conventional in the places I checked ([here](https://hub.docker.com/_/redis) and [here](https://hub.docker.com/_/postgres)), but I kinda like the dotless version smile

Oh yeah they totally are for the tags! I mostly was referring to the extension on the files to indicate the OS. Just the way our CI is currently setup made it easier to just make the OS reference used as the extension the same as the one used for the tag. I also like the undotted one though so let's roll with that for now until we change our minds :smile:

PR #422, opened 2022-11-07
Add `workflow_call` to preview action to specify product version

tylfin:
Our infrastructure using the public preview images to pull from and test changes. This is useful for two reasons:

1. We can specify build versions that may occur between the CRON runs
2. We can specify builds outside the latest, e.g. release candidates, to pin our test infra to during regression

### DIFF
--- a/.github/workflows/build-preview.yaml
+++ b/.github/workflows/build-preview.yaml
@@ -9,10 +9,18 @@ on:
       - dev
       - dev-rspm
   pull_request:
+  workflow_call:
+    inputs:
+      product:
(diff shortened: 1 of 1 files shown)


tylfin:
Oh wait, I think this is already handled here: https://github.com/rstudio/rstudio-docker-products/actions/workflows/build-preview-webhook.yaml

colearendt:
Yes! In fact, Workbench just started triggering specific builds in Jenkins here: https://github.com/rstudio/rstudio-pro/pull/3998

PR #425, opened 2022-11-09
Update rsc jammy image versions

### LINKED ISSUE #424: connect-content-init jammy image version mismatch
We missed a version update somewhere with the 2022.10 release in our content-init jammy image:

```
david@davids-mbp:connect » docker run --rm -it rstudio/rstudio-connect:jammy-2022.10.0--db7dfd1 cat /opt/rstudio-connect/VERSION
2022.10.0
david@davids-mbp:connect » docker run --rm -it rstudio/rstudio-connect-content-init:jammy-2022.10.0--db7dfd1 cat /opt/rstudio-connect-runtime/VERSION
2022.08.1
```

dbkegley:
fixes #424

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -78,7 +78,9 @@ update-rsc-versions:
   sed {{ sed_vars }} "s/RSC_VERSION=.*/RSC_VERSION={{ RSC_VERSION }}/g" \
     connect/.env \
     connect/Dockerfile.bionic \
-    connect-content-init/Dockerfile.bionic
+    connect/Dockerfile.jammy \
+    connect-content-init/Dockerfile.bionic \
+    connect-content-init/Dockerfile.jammy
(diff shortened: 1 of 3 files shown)


ianpittwood (approved):

Justfile
```diff
@@ -96,6 +98,7 @@ update-r-versions:
     package-manager/.env \
     workbench/Dockerfile.bionic \
     connect/Dockerfile.bionic \
+    connect/Dockerfile.jammy \
```
aronatkins, line 101: Looks like `workbench`, `package-manager`, and `r-session-complete` also have `Dockerfile.jammy`.

@ianpittwood -- Is work to add these images still in-flight?

aronatkins (approved):
Looks like Connect is happy, but maybe not the other products.

Justfile
ianpittwood: These should be resolved in #420

PR #426, opened 2022-11-10
allow PATH lookups to be skipped

### LINKED ISSUE #423: Content - examine-images.sh and python 3.6.9
I am creating images for the Kuberentes Connect installation approach, including some R-only images.

In review of the output that is coming out of `content\scripts\examine-image.sh` , I noticed that it was picking up Python 3.6.9 when there was no Python being installed. When I look closer at my pipeline, for an image that is solely installing R, I see this:

`Found Python at /usr/bin/python3 in PATH with canonical location /usr/bin/python3.6 and version 3.6.9.`

I believe python 3.6.9 is getting installed with this chunk in the [Dockerfile.bionic](https://github.com/rstudio/rstudio-docker-products/blob/main/content/base/Dockerfile.bionic#L30).

Assuming that Python 3.6.9 is coming over because of some sort of dependency, is there a better way to adjust the [examine-image.sh](https://github.com/rstudio/rstudio-docker-products/blob/main/content/scripts/examine-image.sh#L162) file to ignore the Python in a `/usr/bin/python3.6` folder?

aronatkins:
Bypasses `PATH` probing for R, Python, and Quarto installations.

fixes #423

### DIFF
--- a/content/README.md
+++ b/content/README.md
@@ -71,6 +71,22 @@ This command generates a single YAML for the available `rstudio/content-pro` ima
 		rstudio/content-pro:r4.1.0-py3.9.2-bionic  > runtime.yaml
 ```
 
+The `NO_PATH_LOOKUPS` environment variable indicates that `PATH` is ignored
+when attempting to discover R, Python, and Quarto installations.
+
+```console
+NO_PATH_LOOKUPS=1 ./scripts/build-image.sh \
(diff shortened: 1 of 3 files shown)


dbkegley (approved):

aronatkins:
The Connect daily builds are erring because we have a problem with our `connect --version` output.
The workbench and r-session-complete dailies are known failures according to @colearendt.

PR #430, opened 2022-11-16
Fix logging config

gbrlsnchs:


aronatkins (commented):
This change will need to be combined with the 2022.11.0 version update. CC @Lytol

Lytol:
Closing in favor of #431

PR #440, opened 2022-11-28
Upgrade pro drivers to 2022.11.0

bschwedler:
Update the default version of the pro-drivers to be `2022.11.0`

As part of this the `update-drivers-versions` just command was updated to update all appropriate files and the `DRIVERS_VERSION_RHEL` variable now derives from the `DRIVERS_VERSION` variable.


melissa-barca (approved):
LGTM!

PR #442, opened 2022-12-07
Bump Package Manager version to 2022.11.4-20

glin:
Update Package Manager to 2022.11.4-20 for another hotfix release.


docker-compose.yml
```diff
@@ -43,8 +43,8 @@ services:
       context: ./package-manager
       dockerfile: "Dockerfile.${IMAGE_OS:-ubuntu1804}"
       args:
-        RSPM_VERSION: 2022.11.2-18
-    image: rstudio/rstudio-package-manager:2022.11.0
+        RSPM_VERSION: 2022.11.4-20
+    image: rstudio/rstudio-package-manager:2022.11.4-20
```
glin, line 47: Are these `image` tag changes in docker-compose.yml intended? I updated `RSPM_VERSION` and ran `just update-versions`.

docker-compose.yml
```diff
@@ -8,7 +8,7 @@ services:
       dockerfile: "Dockerfile.${IMAGE_OS:-ubuntu1804}"
       args:
         RSW_VERSION: 2022.07.2+576.pro12
-    image: rstudio/rstudio-workbench:2022.07.2
+    image: rstudio/rstudio-workbench:2022.07.2-576.pro12
```
colearendt, line 11:
```suggestion
    image: rstudio/rstudio-workbench:2022.07.2
```

docker-compose.yml
colearendt, line 47:
```suggestion
    image: rstudio/rstudio-package-manager:2022.11.4
```

colearendt (approved):
I think this is a bug in the `update-versions` command. We can fix it out of band.

Images now do not have the build number included in their tag, so the compose image reference should not include that

glin:
@colearendt Thanks, the image tags look right now. There are some build failures in the workbench/r-session preview images, but I assume those are unrelated since the same jobs are failing on main. I'll merge when CI passes.

PR #444, opened 2022-12-08
Fix floating license server

ianpittwood:
I encountered some problems running the floating license server for some testing so I'm committing the changes it took for me to get it to run correctly. I also implemented the relevant licensing server variables into the Justfiles so it's easier to run those against the floating license server in the future.

### DIFF
--- a/float/Dockerfile.ubuntu1804
+++ b/float/Dockerfile.ubuntu1804
@@ -6,15 +6,15 @@ ARG DEBIAN_FRONTEND=noninteractive
 RUN apt-get update \
     && apt-get install --no-install-recommends -y \
       ca-certificates \
-	    gdebi-core \
-	    curl \
-	    wget  \
+      gdebi-core \
+      curl \
+      wget \
     && rm -rf /var/lib/apt/lists/*
 
 # Runtime settings
 ARG TINI_VERSION=0.18.0
-RUN curl -L -o /usr/local/bin/tini https://github.com/krallin/tini/releases/download/v${TINI_VERSION}/tini && \
-    chmod +x /usr/local/bin/tini
+RUN curl -L -o /usr/local/bin/tini https://github.com/krallin/tini/releases/download/v${TINI_VERSION}/tini \
+    && chmod +x /usr/local/bin/tini
 
 # install license server binary
 ARG PRODUCT=rsp
--- a/float/Dockerfile.ubuntu1804
+++ b/float/Dockerfile.ubuntu1804
@@ -3,12 +3,13 @@ LABEL maintainer="RStudio Docker <docker@rstudio.com>"
 ARG DEBIAN_FRONTEND=noninteractive
 
 # install deps
-RUN apt-get update && \
-	apt-get install --no-install-recommends -y \
-	  gdebi-core \
-	  curl \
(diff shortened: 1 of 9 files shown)


float/Dockerfile.ubuntu1804
```diff
@@ -3,12 +3,13 @@ LABEL maintainer="RStudio Docker <docker@rstudio.com>"
 ARG DEBIAN_FRONTEND=noninteractive
 
 # install deps
-RUN apt-get update && \
-	apt-get install --no-install-recommends -y \
-	  gdebi-core \
-	  curl \
-	  wget && \
-    rm -rf /var/lib/apt/lists/*
+RUN apt-get update \
```
atheriel, line 6: Are these whitespace changes necessary?

ianpittwood: Not really. I've just been making all the files use the same formatting as I update or change them.

atheriel: That makes sense, then.

float/license-server.conf
```diff
@@ -0,0 +1,20 @@
...
+    <bind port="{{PORT}}"/>
+
+    <!-- # of worker threads (0 means "one per CPU") -->
+    <cpu threads="0"/>
+
+    <!-- license lease time in seconds -->
+    <lease length="1800"/>
+
+    <!-- file to write logs to -->
+    <log file="/var/log/rstudio-licensing.log" level="warning"/>
```
colearendt, line 13: Does the license server have a way to log to stdout or stderr?

float/Dockerfile.ubuntu1804
```diff
@@ -3,53 +3,32 @@ LABEL maintainer="RStudio Docker <docker@rstudio.com>"
...
-    <!-- reactivation settings -->\n\
-    <isgenuine days_between="90" grace="14"/>\n\
-\n\
-    <!-- use proxy -->\n\
-    <!-- <proxy url="http://user:pass@127.0.0.1:8080/"/> -->\n\
-</config>\n\
-\n' \
-| sed -s "s/{{PORT}}/${PORT}/" > /etc/${PRODUCT}-license-server.conf && \
-sed -i "s/\${PRODUCT}/${PRODUCT}/" /usr/local/bin/startup.sh
+RUN sed -i "s/{{PORT}}/${PORT}/" /etc/${PRODUCT}-license-server.conf \
```
colearendt, line 30: Nice! We had this inline to make the Dockerfile more portable, but it needs the custom `startup.sh` too, so it makes sense to keep it separate 😄

colearendt (approved):
Nice! LGTM! Just a couple of questions

float/license-server.conf
ianpittwood: I did yank this verbatim out of the Dockerfile because it wasn't writing correctly in my environment so I'm not sure, but I'll take a look and see if there's anything else we should be doing.

float/Dockerfile.ubuntu1804
ianpittwood: Ah makes sense! My environment was having trouble creating a valid XML file from that echo so that's why I moved it.

PR #454, opened 2022-12-21
Bump Workbench Version to 2022.12

colearendt:
- bump workbench version to `2022.12.0`
- switch workbench and r-session-complete to use and build from product-base-pro images
- remove `install-vs-code` helpers (no longer present)
- add README.md section highlighting how images inherit / relate to one another
- bump to the new daily build line for workbench

Ultimately, we should do some testing with the newly built `dev-` images before we merge this, as there are a fair number of changes here.


colearendt:
Something weird is still broken for azure

colearendt:
Queued up to merge on Tuesday!

colearendt:
Did some rudimentary testing. More to commence, and we will fix forward any issues!

PR #456, opened 2022-12-29
Update deprecated `set-output` commands to use `>> $GITHUB_OUTPUT`

### LINKED ISSUE #418: Fix Github CI warnings for use of deprecated features
We started getting the following warning recently:
```
Warning: The `save-state` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
```
According to the changelog page, the `save-state` and `set-output` commands as we currently use them will be deprecated on May 31st 2023.

[https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)

ianpittwood:
Closes #418 

Github is changing the way their workflow export functions work and are therefore deprecating functions we currently use like `set-output`. This PR updates all uses of `set-output` to `>> $GITHUB_OUTPUT`.

### DIFF
--- a/.github/workflows/build-content.yaml
+++ b/.github/workflows/build-content.yaml
@@ -13,7 +13,7 @@ jobs:
     outputs:
       matrix: ${{ steps.set-matrix.outputs.matrix }}
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - id: set-matrix
         run: |
           MATRIX=$(jq -Mcr < content/matrix.json)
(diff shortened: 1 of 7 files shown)


colearendt (approved):
Good to know! How did you get notification for this? I suspect it will break some other repos too 😞

ianpittwood:
> Good to know! How did you get notification for this? I suspect it will break some other repos too 😞 

As far as I know, you only see them as warnings in the CI logs. We have until June 1st 2023 to fix all these though fortunately.

bdeitte:
Something to consider sending out a PSA on in the New Year when more people are around.  I see a whole bunch of usages: https://github.com/search?q=org%3Arstudio+%22set-output%22&type=code

ianpittwood:
> Something to consider sending out a PSA on in the New Year when more people are around.  I see a whole bunch of usages: https://github.com/search?q=org%3Arstudio+%22set-output%22&type=code

Any idea what the best way to communicate it would be? Folks will also need to update their action versions too I found. The older versions of actions utilize the same deprecated functions.

bdeitte:
Shared in Slack my thoughts on this!

PR #461, opened 2023-01-20
Leaky licensing documentation and workaround

ianpittwood:


connect/README.md
```diff
@@ -150,12 +150,41 @@ docker run -it --privileged \
...
+the license state data across restarts of the container. **These directories differ between products.**
+
+* License Key
+  * `/var/lib/.local`
+  * `/var/lib/.prof`
+  * `/var/lib/rstudio-connect`
+* Floating License
+  * `/var/lib/.TurboFloat`
+
+Please note that the files created in these directories are hardware locked and non-transferable between hosts. Due to
```
kfeinauer, line 178: Excellent explanation!

connect/README.md
```diff
@@ -150,12 +150,41 @@ docker run -it --privileged \
...
+activation state on the container has been lost.
+
+To avoid "leaking" licenses, we encourage users not to force kill containers and to use `--stop-timeout 120` and 
+`--time 120` for `docker run` and `docker stop` commands respectively. This helps ensure the deactivation script has 
+ample time to run properly.
+
+In some situations, it can be difficult or impossible to avoid a hard termination (e.g. power failure, 
+critical error on host). Unfortunately, any of these cases can still cause a license to leak an activation. To help
+prevent a license leak in these situations, users can mount the following directories to persistent storage to preserve
+the license state data across restarts of the container. **These directories differ between products.**
```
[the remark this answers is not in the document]
ianpittwood: There's one product directory that's different for each of the activation key directory sets. Though I guess we could say "may differ" instead. Originally I was going to include the product directories for the floating directory sets as well, but I concluded that they're not actually necessary as they usually don't get written to in that configuration.

Mostly I wanted to ensure that end users read through the README associated with the product they're setting up rather than assuming they're all the same setup.

package-manager/Justfile
```diff
@@ -65,11 +68,32 @@ test-i TAG=`just _make-default-tag` VERSION=RSPM_VERSION:
 run TAG=`just _make-default-tag` CMD="":
   #!/usr/bin/env bash
   set -euxo pipefail
-  if [ -z "{{ RSPM_LICENSE }}" || -z "{{ RSPM_LICENSE_SERVER }}" ]; then
+  if [ -z "{{ RSPM_LICENSE }}" ] && [ -z "{{ RSPM_LICENSE_SERVER }}" ]; then
     echo "Please set RSPM_LICENSE or RSPM_LICENSE_SERVER before running."
     exit 1
   fi
+
+  volume_opts=()
```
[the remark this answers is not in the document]
ianpittwood: That's a fair point. I'll think on if there's a way we could generate these as a target in the root `Justfile`.

float/Dockerfile.ubuntu1804
```diff
@@ -30,6 +30,9 @@ COPY license-server.conf /etc/${PRODUCT}-license-server.conf
 RUN sed -i "s/{{PORT}}/${PORT}/" /etc/${PRODUCT}-license-server.conf \
     && sed -i "s/\${PRODUCT}/${PRODUCT}/" /usr/local/bin/startup.sh
 
+RUN ln -s /var/lib/.yocal /var/lib/.local-float \
```
colearendt, line 33: 😮 `.yocal`, really?

package-manager/README.md
```diff
@@ -152,12 +152,41 @@ docker exec -it {container-id} /bin/bash
...
 There is currently a known licensing bug when using our products in containers. If the container is not stopped
-gracefully, license deactivation may not work properly. To avoid "leaking" licenses, we encourage users not to force
-kill containers and to use `--stop-timeout 120` and `--time 120` for `docker run` and `docker stop` commands
-respectively. This helps ensure the deactivation script has time to run properly. We are still investigating a
+gracefully, the license deactivation step may fail or be skipped. Failing to deactivate the license can result in a 
+"license leak" where a product activation is used up and cannot be deactivated using traditional methods as the 
+activation state on the container has been lost.
+
+To avoid "leaking" licenses, we encourage users not to force kill containers and to use `--stop-timeout 120` and 
+`--time 120` for `docker run` and `docker stop` commands respectively. This helps ensure the deactivation script has
```
colearendt, line 165: Has your testing provided any guidance for how long deactivation "usually" takes? Do we retry at all (i.e. network blips / etc.)?

package-manager/README.md
```diff
@@ -152,12 +152,41 @@ docker exec -it {container-id} /bin/bash
...
+* Floating License
+  * `/home/rstudio-pm/.TurboFloat`
+
+Please note that the files created in these directories are hardware locked and non-transferable between hosts. Due to
+the nature of the hardware fingerprinting algorithm, any low-level changes to the host or container can cause existing
+license state files to invalidate. To avoid this problem, we advise that product containers are gracefully shutdown
+and allowed to deactivate prior to changing any hardware or firmware on the host (e.g. upgrading a network card or 
+updating BIOS) or the container (e.g. changing the network driver used or the allocated number of CPU cores).
+
+While preserving license state data *can* help avoid license leaks across restarts, it's not a guarantee. If you run
```
colearendt, line 186: If you do preserve the license state data, is it possible to deactivate on startup, to make sure the license is cleared? Does activation over a not-properly-deactivated license state clear the state and activation properly?

My suspicion is that it just blindly overwrites and then you still lose the activation

colearendt (commented):
This is awesome work! Well done! Just a handful of questions to help me understand better (and what potential next steps might be).

package-manager/README.md
ianpittwood: For me, I think the longest I've seen it take is around 15 seconds. I'm fairly sure the license-manager binary doesn't retry and nor does our `trap` script. That's a good idea though. The only barrier I see is the license-manager binary doesn't seem to always return a non-zero exit code on failure so it may be difficult to tell if it fails. I guess maybe we could check if the state files have been purged of data? That should be a tell tale sign of deactivation. I'll look into this.

package-manager/README.md
ianpittwood: During my tests, the activation command would skip if the license state files have any data during its read step. It was also able to deactivate the existing license state data across restarts a majority of the time. I'll double check this behavior and make sure we don't need to put any extra logic into the startup script.

ianpittwood: I just tested this with Connect:
1. Confirmed license key was presently at zero activations to start.
2. Started Connect container with persistent license state volumes. Confirmed license key activations ticked to one and that container and software were licensed and working properly.
3. Force killed Connect container. Confirmed license key activations were still at one and that license state files still had activation data.
4. Started Connect container with same persistent volumes containing license state files. Confirmed that license key activations were still at one and that the container and software were still licensed and working.
5. Gracefully stopped container and confirmed that license key activations went back down to zero.

So thankfully it is smart enough not to just blindly overwrite! The same steps should work for Workbench and PPM too.

package-manager/README.md
ianpittwood: I added some retry logic in Connect, PPM, and Workbench! I was correct that license-manager doesn't ever return any other code but zero so I have it checking for any non-empty file in the `.local` directory instead and retrying if there's a file with content in there. I don't foresee a situation where there would be multiple state files in there so it should work well enough.

package-manager/README.md
colearendt: Nice!! Thanks for confirming! I have to say, (4) sounds to me like it deactivated/reactivated if the software was still working / licensed... just because the IP address of the container likely changed?

float/Dockerfile.ubuntu1804
colearendt: Oh no, I think this is good! It just looks like a typo 🤣 Thanks for the background!

colearendt (approved):
Looks awesome!! Well done 😄 

For the record, I have a minor concern around: (1) the license directory not existing, (2) files in that directory not being related to licensing. I talked myself out of those being problems in our containers though 😄

PR #463, opened 2023-01-24
Use new base image for Connect

### LINKED ISSUE #447: Update existing images to use new base images
This is a follow-up to the changes implemented by #415. It added a series of base images and base images with pro drivers for Ubuntu 18.04, Ubuntu 22.04, and CentOS 7 and drafted new Dockerfile definitions for each product that use these new base images. When ready, each of these products should be swapped over to the new base image.

- [ ] Connect
  - [ ] Ubuntu 18.04
  - [ ] Ubuntu 22.04
- [ ] ~~Package Manager~~
  - [ ] ~~Ubuntu 18.04~~
  - [ ] ~~Ubuntu 22.04~~
- [x] r-session-complete
  - [x] Ubuntu 18.04
  - [x] Ubuntu 22.04
  - [x] CentOS 7
- [x] Workbench
  - [x] Ubuntu 18.04
  - [x] Ubuntu 22.04
- [x] Workbench for Azure ML
  - [x] Ubuntu 18.04
  - [x] Ubuntu 22.04

For each image, we need to ensure that it builds, passes existing tests, and functions as expected. The respective `NEWS.md` should also be updated for each product as updates are made.

ianpittwood:
Closes #447 (last one to update)

Replaces existing Dockerfiles using `r-base` base image with drafted Dockerfiles using `product/base` base image.


connect/Dockerfile.ubuntu1804
```diff
@@ -1,115 +1,22 @@
...
-ARG R_VERSION_ALT=4.1.0
-RUN apt-get update -qq && \
-    curl -O https://cdn.rstudio.com/r/ubuntu-1804/pkgs/r-${R_VERSION_ALT}_1_amd64.deb && \
-    DEBIAN_FRONTEND=noninteractive gdebi --non-interactive r-${R_VERSION_ALT}_1_amd64.deb && \
-    rm -f ./r-${R_VERSION_ALT}_1_amd64.deb \
-    && apt-get clean \
-    && rm -rf /var/lib/apt/lists/*
-
-# Install Python  -------------------------------------------------------------#
+ARG R_VERSION=4.2.0
```
aronatkins, line 1: Could you explain what is changing about the contents of the final image?

For example, it looks as if this image is shifting from R-3.6.2 (default) and R-4.1.0 to R-4.2.0 (default) with R-3.6.2. Just this version adjustment by itself deserves a NEWS mention.

Do any other changes come in with the shift to rstudio/product-base?

ianpittwood: Oops I think I typo'd the primary R version in the file so I'll fix that. We can revert the default versions being reversed in the Dockerfile if you foresee it being an issue. Originally I was going to align the version order used for our Python and R versions (latest release first, older release as alt), but I decided it was too large of a change to make on top of all the other base image changes. While the order is changed for the argument defaults in this Dockerfile, [the CI still builds with 3.6.2 as the default and 4.1.0 as the alternative version](https://github.com/rstudio/rstudio-docker-products/blob/main/ci.Justfile#L6-L10) and so does the Connect Justfile. I would eventually like to see the order of the versions be uniform, but it's not very high priority of course. We should add a NEWS entry though, I'll make sure to add that in shortly.

The main change is that both Workbench (which already has this change) and Connect will now use the [product-base image](https://github.com/rstudio/rstudio-docker-products/blob/main/product/base/Dockerfile.ubuntu1804) as their base image. product-base is the combined common dependencies of Workbench and Connect (2 versions of R, 2 versions of Python, etc.) that we made to help reduce the repeated code in the repo and hopefully make maintenance and optimization easier in the future when we need to do it. I'm open to feedback on it if you have concerns about what's included or if something is included unnecessarily!

aronatkins: No explicit concerns; I wanted to have you tell us the changes to the final image so that each reviewer can see what's expected to change without requiring each of us to build the old/new versions and compare.

Thanks for the NEWS entry; if the final contents of the image have substantially changed, that should be given, well Indicating the change in FROM doesn't help the reader understand the impact in contents.

To restate: I'm not questioning the change, only asking for details about how the new image is different.

ianpittwood: Sure, I'll do my best to sum it up.

Package changes:
```diff
- bzip2
+ cmake
+ curl
+ default-jdk
+ dirmngr
+ g++
+ gcc
- gdebi-core
+ gfortran
+ gpg
+ gpg-agent
+ libev-dev
- libglib2.0-0
- libnss-sss
- libpam-sss
+ libopenblas-dev
+ libpaper-utils
+ libpcre2-dev
- libsm6
- libssl1.0.0
- libxext6
- libxrender1
+ unzip
+ zip
```

Other changes:
* `tini` is now installed from the Posit CDN and does file verification.
*  Quarto is installed by default.
* `apt-get` is now used for package installation to replace `gdebi`.
* `wget` is fully replaced by `curl`.
* Python is installed using Posit's own `python-builds`, replacing Miniconda.
* ~300MB size reduction.

aronatkins (approved):
With a Quarto install, we should update the gcfg to enable Quarto support. Doesn't need to occur in this PR, but good as a follow-up.

connect/Dockerfile.ubuntu1804
colearendt: This is a helpful diff! I think `libnss-sss` and `libpam-sss` need to be included - we had a customer PR them a while back in order to work with a sidecar sssd. I am concerned that `libglib2` and `bzip2` may be R package dependencies... do you have any idea where some of these deps being removed came from?

colearendt (commented):
Just one sysdep question - I would like to be careful that someone pulling this image would suddenly lose sysdeps (esp. if R or python packages depend on them). If this is the case, it may be best to couple this change with a bump in Connect's version.

I was also curious if we have figured out our guidance for Quarto version being included and how we plan to update that? @aronatkins does bumping the quarto version require content to be rebuilt? Have changes there been thought through from Connect's point of view?

This lack of lifecycle understanding has been the main hold-up in us adding Quarto to several of the different images (content images, Connect, etc.).

aronatkins:
Connect treats changes to the configured set of Quarto installations similarly to changes to the R or Python installations; the content will need to be re-matched and re-built, which happens automatically, but does have a cost.

ianpittwood:
@colearendt @aronatkins how do we want to proceed with this then? Should we try to devise a better way of handling Quarto in harmony with `content`?

aronatkins:
I think I agree with Cole - we should couple changes to system libraries with a Connect release and flag it as potentially breaking.

In terms of Quarto releases - this image should probably choose some version and stick with it, then handle version updates just like we do for R/Python. 

The `content` images are only used when folks run off-host, while the Quarto/R/Python install in this image is used both with off-host and "native" execution.

ianpittwood:
Sorry I forgot to respond here. I moved #447 to blocked pending the changes from #483 and the next Connect release.

connect/Dockerfile.ubuntu1804
ianpittwood: I added the packages noted by @colearendt and flagged this as potentially breaking.

ianpittwood:
Blocked by #567

msarahan (approved):

bschwedler (approved):

ianpittwood:
Blocked by #568 since we can't pass tests until we have the new base image. Implementing `dev`/`preview` builds of the product base images should be easy enough that we can get it done before EoW for this PR and the eventual release of Connect 2023.06.0 to `main`.

PR #464, opened 2023-01-24
Use new base image for Package Manager

ianpittwood:
Replaces existing Dockerfiles using r-base base image with drafted Dockerfiles using product/base base image.


package-manager/Dockerfile.ubuntu1804
```diff
@@ -1,40 +1,28 @@
-ARG R_VERSION=3.6.2
-FROM rstudio/r-base:${R_VERSION}-bionic
+ARG R_VERSION=4.2.0
+ARG R_VERSION_ALT=3.6.2
+ARG PYTHON_VERSION=3.9.5
+ARG PYTHON_VERSION_ALT=3.8.10
+ARG REGISTRY=ghcr.io
+FROM ${REGISTRY}/rstudio/product-base:ubuntu1804-r${R_VERSION}_${R_VERSION_ALT}-py${PYTHON_VERSION}_${PYTHON_VERSION_ALT}
```
glin, line 6: I'm missing context on the new base images, so just catching up by skimming through https://github.com/rstudio/rstudio-docker-products/pull/415 now. It seems like the new base image has a lot of things installed that weren't here previously, and probably won't be used in a typical Package Manager setup.

For example, PPM really only needs one version of R (and the version doesn't really matter) for Git building, and doesn't need to install R or Python packages either. So the additional system libs, Python, and Quarto will all probably be unused unless a customer is also running other products in this image.

This might be okay, but I wonder if existing users would miss the smaller images. Do you know how much this changes the size of the image? If it's significant enough, I think we should at least document the change in the NEWS.

ianpittwood: Yes, that's all fair points. Package Manager might actually be the only product which it would make sense to use `r-base` as a base image since it has very few requirements compared to the other images. PM was the only image that would be inflated by this change as well, 1.19GB to 2.87GB without compression, so I agree it may be something we should discuss further. Our original intention was to get all of the images on a more comprehensive base image within the repo so it would be easier for us to maintain and optimize in the future. While I still think those changes will work for Connect and Workbench, I'm not sure how we would optimize PM on this base image to anywhere near its original size without just removing stuff we've already installed.

@colearendt what are your thoughts? I know we've discussed previously that most customers prefer the large comprehensive images, but I'm not sure that's the case here.

ianpittwood: OK I'm good with sticking with that for the moment then! Should we close this PR? Or should I just modify it to remove the drafts and revert the images to the old definitions?

glin: Would it make sense to split out the product base image into a more minimal, R-only base image that can be used here? Just for consistency in case `rstudio/r-base` diverges from the base images here (although that image doesn't change often). Or is that too many images to manage?

And just thinking more about that now, I think even `rstudio/r-base` has extra stuff that Package Manager doesn't need, like TinyTeX and Pandoc. We could exclude both of those from a new base image to shave off a few hundred MBs more.

ianpittwood: @glin I think that's something worth considering, but I'm also concerned about overloading ourselves with base images. The only things PPM has in common with the other product images is 1 version of R, `tini`, and a handful of `apt` operations and utilities so I'm not sure if it makes sense trying to force it to share a base image. Perhaps we could build PPM directly from Ubuntu's image? Of course if we're concerned about building directly from upstream I think the next best option would be to look into breaking up the product-base image a little more to accommodate PPM without bloating it too much.

glin: Yep, that makes sense. Building PPM from the Ubuntu base image sounds like a good compromise if the base image would've only been directly used by PPM anyway, and would let us remove the unused TinyTeX and Pandoc. But I'd be fine with either that or staying on `r-base`, and will leave it up to you and Cole.

ianpittwood: @colearendt thoughts?

colearendt: I'm not super opinionated. I'm inclined to let it ride for now - we just did a bunch of changes switching RSPM to rootless. Could be nice to let it sit for a bit 😄 (I.e. users who may have been burned can get some stability for a while)

ianpittwood:
Closing this PR for now and making a note to come back and clean up the drafts in the future and possibly switch PPM straight to the Ubuntu image as its base.

PR #471, opened 2023-02-08
Remove not supported statement as discussed internally

bdeitte:


ianpittwood (changes requested):
We'll want to extend these changes to the READMEs located in the product directories too as they have the same "not formally supported" snippet in there. We may also want to get #461 merged prior to removing the warnings? There's a small change I know I need to make with the Justfiles that Karl requested, but it should otherwise be ready for review.

bdeitte:
@ianpittwood Yes I forgot this was in the other READMEs, good call and doing that now.  And I will wait for the #461 merge

README.md
```diff
@@ -4,17 +4,17 @@ Docker images for RStudio Professional Products
 
 **IMPORTANT:** There are a few things you need to know before using these images:
 
-1. These images are provided as a convenience to RStudio customers and are not formally supported by RStudio. If you
-   have questions about these images, you can ask them in the issues in the repository or to your support
-   representative, who will route them appropriately.
+1. This image may introduce **BREAKING** changes; as such we recommend:
+   - Avoid using the `latest` or `{operating-system}` tags to avoid unexpected version changes, and
+   - Always read through the NEWS.md file in the appropriate sub-folder to understand the changes before updating.
```
bdeitte, line 9: I removed the "not supported" statement but also added the BREAKING statement, felt appropriate to have this here as well while changing to mention sub-folders.

connect/README.md
```diff
@@ -27,12 +27,9 @@ https://www.rstudio.com/products/connect/.
...
 # Notice for support
 
-1. This image may introduce **BREAKING** changes, as such we recommend:
+1. This image may introduce **BREAKING** changes; as such we recommend:
    - Avoid using the `latest` or `{operating-system}` tags to avoid unexpected version upgrades, and
-   - Always read through the [NEWS](./NEWS.md) to understand these changes before updating.
-1. These images are provided as a convenience to RStudio customers and are not yet formally supported by RStudio. If you
-   have questions about these images, you can ask them in the issues in the repository or to your support
-   representative, who will route them appropriately.
+   - Always read through the [NEWS](./NEWS.md) to understand the changes before updating.
```
bdeitte, line 32: Removed "not supported" in all places, and some nit-picky grammar updates.

workbench-for-microsoft-azure-ml/README.md
```diff
@@ -1,5 +1,5 @@
 # RStudio Workbench for Microsoft Azure ML
 
-This image is built specifically for use with Azure ML.
+This image is built specifically for use with Azure ML. Please contact Posit support in order to use.
```
bdeitte, line 3: As I looked at the READMEs, realizing we should add this given a separate license and some more info is needed to use, don't want anyone thinking they can just directly use this on Azure ML without help.

ianpittwood (approved):
LGTM! I just finished the updates to #461 as well so that's ready for reviews.

README.md
colearendt: This reminds me that we should delete that tag for sure, since it is the default.

bdeitte: Oh didn't know that, will update!

Issue #455, opened 2022-12-22
Add opencontainer labels to the docker images


colearendt:

ianpittwood:
https://github.com/opencontainers/image-spec/blob/main/annotations.md

PR #477, opened 2023-02-14
propose new labels and data format

colearendt:
I propose adding new labels to our images, and a general data format of `[{"version": "1.2.3", "path": "/path/to/install", "label": "some-label"}]` for installed software under a tag `versions`.

```
# opencontainer labels
org.opencontainers.image.source = "https://github.com/rstudio/rstudio-docker-products"
org.opencontainers.image.description = "Description of the image"
org.opencontainers.image.title = "Title of the image"
org.opencontainers.image.url = "some-url.com"

# posit labels
posit.os = "ubuntu1804"
posit.quarto.included=true
posit.quarto.versions=[{versiondata}]
posit.python.included=true
posit.python.versions=[{versiondata}, {otherversiondata}]
posit.r.included = true
posit.r.versions = [{versiondata}, {otherversiondata}]
posit.tini.included=true
posit.tini.versions = [{versiondata}]

# other proposals
posit.drivers.included=true
posit.drivers.versions=[{versiondata}]
posit.product.name = "Connect"
posit.product.component = "Server" or "Runtime"
posit.product.version = "2023.01"

posit.workbench.editors...? Some way to put the editors into an array? Path to the editors? Is that overkill?
```

These would replace the current `posit.r.version`, `posit.r.version_alt`, etc.

Labels are articulated here, and seem to me like they could be a useful convention for us to explore as a way of both documenting our images and potentially integrating more smoothly into the products. Azure uses this approach as an interface into their platform, for instance.

https://github.com/rstudio/rstudio-docker-products/blob/ff278dbeaa93503c3ab61024374e78e3cb993b49/workbench-for-microsoft-azure-ml/Dockerfile.ubuntu2204#L163-L174

https://docs.docker.com/engine/reference/builder/#label


dbkegley:
I like this; it could be very helpful on the Connect side when we get to image "discovery".  We could search image labels within a registry to find compatible R/python/quarto versions without having to pull and run every image.

product/base/Dockerfile.ubuntu1804
```diff
@@ -140,5 +140,19 @@ LABEL posit.r.version="${R_VERSION}" \
       posit.python.version_alt="${PYTHON_VERSION_ALT}" \
       posit.tini.version="${TINI_VERSION}" \
       posit.quarto.version="${QUARTO_VERSION}"
+LABEL org.opencontainers.image.source = "https://github.com/rstudio/rstudio-docker-products" \
```
ianpittwood, line 143: nit: we can probably make this a single `LABEL` block if we want to keep the above labels, but I also wouldn't be opposed to replacing them completely since I doubt they have been heavily used yet

ianpittwood (approved):
LGTM!

ianpittwood:
Would this close #455?

meatballhat:
**BUMP** ✋🏼 please may I have this?

atheriel:
@meatballhat I don't know why this conversation stalled but it would be good to confirm that we still actually want this to happen before working on it.

(No objections from me, though -- it sounds like it could eventually help on the discovery side in Connect/Workbench, which is cool.)

bschwedler:
@ianpittwood I think it could be useful to pull in all or a subset of the labels we have been working with in the build tooling update.

bschwedler:
This work will be completed in the in-progress image build refactor.  The work is being tracked in the `posit-dev` GitHub org.

PR #479, opened 2023-02-17
Add file command for rsession-complete images

jeffvroom:
The rsession will try to run `file ...` when trying to open a file without a known suffix.  If that command gets an error, it reports the file as binary so we need this extra command.


melissa-barca (approved):
LGTM

ianpittwood (approved):

cm421 (approved):
LGTM!

r-session-complete/Dockerfile.ubuntu2204
```diff
@@ -27,6 +27,7 @@ RUN apt-get update \
       libpq-dev \
       rrdtool \
       subversion \
+      file
```
ianpittwood, line 30:
```suggestion
      file \
```

ianpittwood (changes requested):
My bad, we missed one `\` here. Other than that I think it should be good pending the builds.

colearendt (approved):
LGTM! Thanks @jeffvroom ! This is an interesting problem - have y'all thought about whether these types of dependencies should be declared explicitly (i.e. as a build / install dependency?). It's weird because it isn't a proper system dependency, but we do shell out to something that we depend on on the system 🤷‍♂️

jeffvroom:
@colearendt Thanks, that's a great suggestion. How about if I cancel this PR and fix this problem in:  https://github.com/rstudio/rstudio-pro/issues/4455?

PR #486, opened 2023-03-08
Add python to the PATH variable in r-session-complete

### LINKED ISSUE #481: Include python in path for r-session-complete
In the r-session-complete images R is on the path by default. Python is not.

Run the r-session-complete image:

```bash
docker run -it --rm rstudio/r-session-complete:jammy /bin/bash
```

In the container:

```bash
root@876544b0ee67:/# which python
root@876544b0ee67:/# which R
/usr/local/bin/R
```

I think it would be helpful for Python users to have the intended default version of Python included on the path. Then, when a python user opens a new session of any kind, when they execute the python command it will point to the intended interpreter. This will also make things more consistent between R and Python, since we are intentionally putting R on the path.

We could also consider configuring VS code to use this interpreter as the default?

ianpittwood:
Closes #481 

I've noticed this issue crop up a few times in the last week with both internal teams and a customer. I prepended the `/opt/python/${PYTHON_VERSION}/bin` path to the `PATH` variable in the container to make it more immediately accessible to users.


AlexMapley (approved):
Nice

melissa-barca (approved):
LGTM

mcbex (approved):
👍

AlexMapley:
I tested this locally and verified it worked with https://github.com/rstudio/platform-helm/pull/60/files 🔥

PR #492, opened 2023-03-14
Add ubuntu2204 content images

ianpittwood:


content/matrix.json
colearendt, line 1: This "EOL'd python versions" problem is an interesting one, and something we should raise w/ the Connect team. I'm not sure they have thought about how to deprecate / stop building / stop supporting these images formally...

dbkegley: I think our preference for this PR would be to only add the new jammy images rather than trying to deprecate the old python/R versions at the same time. We have customers who are currently using those older bionic images and if we remove them now then their content will start to break. Our plan is to move everything over to jammy over the next month or so anyway, and that seems like a good cut-over that will allow us to deprecate bionic and at the same time we can remove the old r/py versions from the matrix.

Not in this PR, but would we also consider an additional docker tag that doesn't include the patch version? This way customers can reference `content-base:r4.1-py3.10` without needing to update every time we take a new patch without changing the minor version? This would be particularly useful for connects helm chart where we define our default set of execution environments for off-host execution. 

cc @aronatkins

aronatkins: @dbkegley - unfortunately, it would mean a full rebuild, today. we require exact version matching when determining if an existing cache is reusable.

dbkegley: dang. I was hoping we would rely on the version matching strategy. ah well, never mind then!

ianpittwood: Dang, I liked that idea too! I'll remove the changes to the Python/R versions for 18.04 if we'll just be getting rid of them at the next release. Are the 22.04 Python/R versions acceptable?

.github/workflows/build-content.yaml
```diff
@@ -117,6 +117,12 @@ jobs:
           docker push rstudio/content-base:r${{ matrix.config.r }}-py${{ matrix.config.py }}-${{ matrix.config.os_alt }}
           docker push ghcr.io/rstudio/content-base:r${{ matrix.config.r }}-py${{ matrix.config.py }}-${{ matrix.config.os_alt }}
 
+      - name: Push image(s)
+        if: ${{ github.ref != 'refs/heads/main' }}
+        run: |  # Push images to ghcr on non-main branches, otherwise Pro images may not build
+          docker push ghcr.io/rstudio/content-base:r${{ matrix.config.r }}-py${{ matrix.config.py }}-${{ matrix.config.os }}
+          docker push ghcr.io/rstudio/content-base:r${{ matrix.config.r }}-py${{ matrix.config.py }}-${{ matrix.config.os_alt }}
+
```
dbkegley, line 125: Is this change needed? It looks like we're pushing the same 2 images again that we just pushed on L117 and L118

content/matrix.json
```diff
@@ -1,20 +1,15 @@
...
+  {"r": "4.1.0", "py": "3.8.16", "drivers": "2022.11.0", "os": "ubuntu1804", "os_alt": "bionic"},
+  {"r": "4.1.0", "py": "3.9.16", "drivers": "2022.11.0", "os": "ubuntu1804", "os_alt": "bionic"},
+  {"r": "4.1.3", "py": "3.10.10", "drivers": "2022.11.0", "os": "ubuntu1804", "os_alt": "bionic"},
+  {"r": "4.2.2", "py": "3.10.10", "drivers": "2022.11.0", "os": "ubuntu1804", "os_alt": "bionic"},
+  {"r": "4.0.5", "py": "3.8.16", "drivers": "2022.11.0", "os": "ubuntu2204", "os_alt": "jammy"},
+  {"r": "4.0.5", "py": "3.9.16", "drivers": "2022.11.0", "os": "ubuntu2204", "os_alt": "jammy"},
+  {"r": "4.1.0", "py": "3.8.16", "drivers": "2022.11.0", "os": "ubuntu2204", "os_alt": "jammy"},
+  {"r": "4.1.0", "py": "3.9.16", "drivers": "2022.11.0", "os": "ubuntu2204", "os_alt": "jammy"},
+  {"r": "4.1.3", "py": "3.10.10", "drivers": "2022.11.0", "os": "ubuntu2204", "os_alt": "jammy"},
+  {"r": "4.2.2", "py": "3.10.10", "drivers": "2022.11.0", "os": "ubuntu2204", "os_alt": "jammy"}
```
dbkegley, line 14: The only thing that I would like to see added is an R 3.6.3 for jammy. 

I was planning to go even more minimal for this matrix originally. I was thinking we could have 4 images, with the last 4 minor versions of both R and Python.

```json
  {"r": "3.6.3", "py": "3.8.16", "drivers": "2022.11.0", "os": "ubuntu2204", "os_alt": "jammy"},
  {"r": "4.0.5", "py": "3.9.16", "drivers": "2022.11.0", "os": "ubuntu2204", "os_alt": "jammy"},
  {"r": "4.1.3", "py": "3.10.11", "drivers": "2022.11.0", "os": "ubuntu2204", "os_alt": "jammy"},
  {"r": "4.2.1", "py": "3.11.3", "drivers": "2022.11.0", "os": "ubuntu2204", "os_alt": "jammy"}
```

dbkegley (approved):

.github/workflows/build-content.yaml
colearendt: Wait why are we pushing on non-main...? That seems concerning

ianpittwood: We have to push new base images somewhere so they can be used in the next stage. This block pushes only to GHCR on non-main builds so we can still build both the base and pro versions of the image. The other block pushes to Docker Hub and GHCR on main builds. I usually hold off on pushing to Docker Hub outside of main builds since it's a little more publicly visible.

ianpittwood: That's the way that the product base works as well, but if we're concerned I can see if we can cache and restore the tarball of the image. `buildx` is just kind of a pain in these multi-stage builds unfortunately.

ianpittwood: I removed the push steps so we could just merge this when ready if you're good with it @colearendt @dbkegley. In the future we can try to fix the multi-stage build problems in another PR.

dbkegley:
LGTM, thanks you for the quick turnaround, @ianpittwood !

.github/workflows/build-content.yaml
colearendt: Ahh interesting. I didn't realize that was the case 😬 That's super sketchy IMO. Definitely up for taking another pass to resolve 😄

colearendt (approved):
LGTM! Nice!

.github/workflows/build-content.yaml
ianpittwood: @colearendt please let me know if you have any secret tricks for reusing local images as parent images in `buildx`! Docker's support around still seems shoddy at best and falls way short of Docker "classic" and my research hasn't given me anything that works unfortunately. I know Podman can do it, but I'm not sure if we want a solution that involves switching our builder entirely.

PR #499, opened 2023-03-21
Workbench 2023.03.0

skyeturriff:
Bump release version to 2023.03.0+386.pro1

Small update to `sed` expression so it would catch versions in READMEs


colearendt:
@ianpittwood did we want to batch any R / python version changes into this change of Workbench?

colearendt:
Nice! Yeah, I think `dev-workbench` is a great target. Or you can use this path for sure, too 😄

cm421 (approved):
LGTM

colearendt (approved):
LGTM!

PR #507, opened 2023-03-27
Switch and upgrade R and Python versions for Workbench

ianpittwood:
Swaps the order of the R versions (primary now being the latest version) and upgrades the R and Python versions to later releases/patch versions. This did require some modifications to the build pipeline in order to build with the correct python/R versions only on Workbench images.


colearendt:
This generally looks awesome! Just a few issues passing params through, I think

ianpittwood:
Alright @colearendt I'm fairly confident I hammered out all the problems. Ready for review!

colearendt (approved):
LGTM! I'll go ahead and merge this so that we can review the other branch in the AM 😄

PR #534, opened 2023-04-28
stop pushing on non-main branches

colearendt:
If we want to follow up on a "better" pattern for building the full pipeline on test branches, we should take that as a separate item. The fact that random branches can overwrite prod images without warning should be removed.


bdeitte (approved):
Ian is out today, but looking back I believe this is just a mistake when testing things and not meant to be merged.  Definitely makes sense to switch back, thanks.

ianpittwood (approved):
Yup, LGTM. This was more to get around a bug with `buildx` being unable to reuse local images as parent images for these multi-stage buildx. We should try to fix that at some point, but thankfully it only will come up on references to parent images with new tags until then.

Issue #305, opened 2022-05-04
Evaluate operating system choices


colearendt:
- Bionic is scheduled for EOL April 2023
- Our OS choices impact our customers, because we only build bionic (for now). And CentOS7 for r-session-complete
- We have had many requests for CentOS / RHEL (i.e. #111 , #225 , in several customer calls, etc.)
- Ubuntu Jammy Jellyfish has been released, which has 

We should decide:
- do we have enough collective buy-in / dev effort to start forking the build matrix and maintaining different distros? How many? For which products? (i.e. RSPM is arguably impacted least / is most OS agnostic)
- at what point should we pull the trigger on stopping bionic builds? 
- when should the `latest` / `versioned` images change off of bionic
- tag paradigms to make these types of differences clearer / cleaner? (related to #302  )
- #255 so that the number of versions we build is minimized in the face of more OS's
- desired product features / documentation to make migration across OS easier
    - i.e. RSC with Launcher arguably makes the OS switch for the "main" service less of an issue
    - finding a way to build `r-session-complete` less often (i.e. more like the `content` images) could simplify effort considerably or merge the image with `content` for RSC. However, this is likely to take a _long_ time for customers to pick up...
- how to stage this change to minimize impact (i.e. couple with a versioned release, etc.)

Notes:
- https://hub.docker.com/r/redhat/ubi8 (for RHEL)
- https://hub.docker.com/_/ubuntu (`ubuntu:jammy`)
- Not all images need to move at the same time. But we should explain / prepare customers appropriately so they can plan. There could be improvements in our notification streams / channels, etc.

bdeitte:
We should consider at least talking to AWS and AzureML about our options when this gets further.

mbuet2ner:
Hey @colearendt,

Thanks for beeing transparent on the future of the base system(s) and involving the community!

Are there any updates on the subject and if not do you maybe have a rough timeline, when a decision is going to be made?
It would be especially interesting *which* CentOS sucessor is picked – if any.

bdeitte:
[...]

PR #540, opened 2023-05-10
Start Preview and Daily builds on Focal

AndrewMcClain:
Goals Here:
Build an Ubuntu 2004 Focal base image
Switch our preview and daily builds to focal
Still have to maintain 1804 Bionic until 2023.05 released, so latest still produces bionic builds as of now.

On 2023.05's release, we'll update this again to pull the remaining latest build onto focal.

### DIFF
--- a/.github/workflows/build-latest.yaml
+++ b/.github/workflows/build-latest.yaml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ['ubuntu1804', 'ubuntu2204', 'centos7']
+        os: ['ubuntu1804', 'ubuntu2004', 'ubuntu2204', 'centos7']
         lang:
           - {r-primary: "4.2.0", r-alternate: "3.6.2", py-primary: "3.9.5", py-alternate: "3.8.10"}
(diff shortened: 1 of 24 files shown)


cm421:
To get the `product-base-pro` focal images in place, should the `product/base/Dockerfile.ubuntu2004` dockerfile be in its own PR? Once that's in, then these automated checks should pass

colearendt:
Wait, why are we switching to Focal? We are already building Jammy images?

colearendt:
> This mirrors the bionic/jammy distinction we had previously.

I personally don't think it's worth the effort. The reason we are maintaining both `bionic` and `jammy` right now is b/c `bionic` is EOL.  We "skipped" focal, so I'd save the development effort for a discussion about whether to support RHEL 8 or RHEL 9 images (#305), and perhaps another ubuntu LTS line next year (24.04). 

If we choose to maintain an image, we should do it for some customer value / reason. At present, focal has the smallest potential upside of any new OS choice IMO. Happy to be persuaded otherwise, but I'd suggest we focus effort elsewhere and make such OS choices carefully / platform wide (since it forks our entire development line). We also have latent demand for ARM images too.

AndrewMcClain:
After Discussion with Maria, Cole and Zach, the introduction of focal images is unnecessary, so I'm closing this PR as won't merge. I'll open a new PR with the changes to just stop building Bionic images.

PR #543, opened 2023-05-12
Bump Workbench to CB version 2023.03.1

zachhannum:


samssann:
any update when this will be released? looking forward to the newest version which will hopefully solve some kubernetes related problems we have been having (rstudio/launcher#604 and rstudio/launcher#645)

tuge98:
Hopefully this will be released soon!

ianpittwood (changes requested):
LGTM, but I think we missed a few spots:
- `docker-compose.yml` still contains a reference to the previous version tag of the image.
- `NEWS.md` entries should be updated for the projects.
- The `README.md` files in each project have a list of "supported tags" that don't look like they've been updated. I'd appreciate if you could tick those up as well! They're automatically published to Docker Hub in addition to Github so they're a helpful reference for customers.

colearendt (approved):
LGTM!

PR #545, opened 2023-05-19
Remove "dev-" prefixing for preview/daily builds

ianpittwood:
I spoke with the IDE team this morning and we determined the "dev-" prefix was causing some confusion over what branch was being used after the branch restructure. Since really all they need is the suffix version, I think we can completely remove the "dev-" prefixes for preview builds.

### DIFF
--- a/ci.Justfile
+++ b/ci.Justfile
@@ -176,13 +176,6 @@ build-preview $TYPE $PRODUCT $OS $VERSION $BRANCH=`git branch --show`:
   TAG_CLEAN_VERSION=`just _get-clean-version $VERSION`
   TAG_VERSION=`just _get-tag-safe-version $VERSION`
 
-  # set branch prefix
-  if [[ $BRANCH == "dev" ]]; then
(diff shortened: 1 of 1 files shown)


AndrewMcClain (approved):
LGTM!

astayleraz (approved):
LGTM

ianpittwood:
Hold on, rethinking if this fully solves the problem...

.github/workflows/build-preview.yaml
```diff
@@ -50,8 +51,14 @@ jobs:
         with:
           ref: 'main'
 
+      - name: Check Out dev Branch
```
colearendt, line 54: We actually don't need this. The default is to checkout `dev` 😄 (since the cron runs on `dev`). Just the line 7 change should suffice

ianpittwood: LOL good point, I'll fix that :laughing:

colearendt (approved):
LGTM!

PR #557, opened 2023-06-08
Bump PWB version to 2023.03.2+454.pro2

cm421:
Note I didn't update the NEWS.md files, because there was no changes outside the PWB version bump. Is that fine or do we prefer an entry like in `workbench/NEWS.md`:

> 2023.03.1
   No changes


zachhannum (approved):
LGTM! FWIW, the only reason I added the "No Changes" entry was because I actually had a news entry for session.

ianpittwood (approved):
LGTM!

PR #564, opened 2023-06-22
Add Workbench for Google Cloud Workstations image

ianpittwood:
Adds the image definition for the initial release of Workbench for Google Cloud Workstations


bschwedler (changes requested):
Most of my comments are geared toward attempting to reduce the size of the image layers where possible.  

I may have missed some places where we can do this, especially if they were buried in a script.
Also, is there an R package cache we can remove after packages are installed like I suggested for `pip`?  I don't know as much about managing packages there.

Note: I did not do a thorough review of the configuration files

workbench-for-google-cloud-workstations/.env
```diff
@@ -0,0 +1,10 @@
+RSW_VERSION=2023.06.0+421.pro1
+RSW_DOWNLOAD_URL=https://download2.rstudio.org/server/focal/amd64
+RSW_NAME=rstudio-workbench
+PYTHON_VERSION=3.10.11
+PYTHON_VERSION_ALT=3.9.16
+PYTHON_VERSION_JUPYTER=3.9.16
+R_VERSION_ALT=4.1.3
+R_VERSION=4.2.3
+DRIVERS_VERSION=2023.05.0
+QUARTO_VERSION=1.3.324
```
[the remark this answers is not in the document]
ianpittwood: Good idea. I believe I've condensed everything down to the `.env` file now. This won't match up with the other projects in the repo, but it's a good pattern to try out and see if we want to update the other projects to follow it.

workbench-for-google-cloud-workstations/README.md
```diff
@@ -0,0 +1,279 @@
...
+
+* Maintained by: [the Posit Docker team](https://github.com/rstudio/rstudio-docker-products)
+* Where to get help: [our Github Issues page](https://github.com/rstudio/rstudio-docker-products/issues)
+* RStudio Workbench image: [Docker Hub](https://hub.docker.com/r/rstudio/rstudio-workbench)
+* RStudio r-session-complete image: [Docker Hub](https://hub.docker.com/r/rstudio/r-session-complete)
+
+# Supported tags and respective Dockerfile links
+
+* [`2022.07.2`, `bionic`, `ubuntu1804`, `bionic-2022.07.2`, `ubuntu1804-2022.07.2`](https://github.com/rstudio/rstudio-docker-products/blob/main/workbench/Dockerfile.ubuntu1804)
+* [`jammy`, `ubuntu2204`, `jammy-2022.07.2`, `ubuntu2204-2022.07.2`](https://github.com/rstudio/rstudio-docker-products/blob/main/workbench/Dockerfile.ubuntu2204)
```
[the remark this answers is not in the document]
ianpittwood: I'm actually going to remove most of this file and link to Google Cloud Workstations docs instead. Most of the information in here won't apply to Workstations as it would to the normal image (where it was copied from).

workbench-for-google-cloud-workstations/README.md
```diff
@@ -0,0 +1,279 @@
...
+
+```
+# Replace with valid license
+export RSW_LICENSE=XXXX-XXXX-XXXX-XXXX-XXXX-XXXX-XXXX
+
+# Run without persistent data using default configuration
+docker run -it \
+    -p 8787:8787 \
+    -e RSW_LICENSE=$RSW_LICENSE \
+    rstudio/rstudio-workbench:ubuntu1804
```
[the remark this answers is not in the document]
ianpittwood: Same as above.

workbench-for-google-cloud-workstations/Dockerfile
```diff
@@ -0,0 +1,197 @@
...
+       tar xzvf - --strip 2 -C /opt/rstudio-license/ \
+    && chmod 0755 /opt/rstudio-license/license-manager \
+    && mv /opt/rstudio-license/license-manager /opt/rstudio-license/license-manager-orig \
+    && rm -f /usr/lib/rstudio-server/bin/license-manager
+
+COPY --chmod=755 license-manager-shim /opt/rstudio-license/license-manager
+COPY --chmod=0775 startup.sh /usr/local/bin/startup.sh
+COPY startup-launcher/* /startup/launcher/
+COPY startup-user-provisioning/* /startup/user-provisioning/
+COPY startup/* /startup/base/
```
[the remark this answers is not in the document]
ianpittwood: I combined them into a single block and tested to make sure things still worked.

workbench-for-google-cloud-workstations/Justfile
```diff
@@ -0,0 +1,100 @@
...
+run TAG=`just _make-default-tag` CMD="":
+  #!/usr/bin/env bash
+  set -euxo pipefail
+  if [ -z "{{ RSW_LICENSE }}" ] && [ -z "{{ RSW_LICENSE_SERVER }}" ]; then
+    echo "Please set RSW_LICENSE or RSW_LICENSE_SERVER before running."
+    exit 1
+  fi
+
+  docker run -it --privileged \
+    ${volume_opts[@]} \
```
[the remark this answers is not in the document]
ianpittwood: Removed `-x` flag

workbench-for-google-cloud-workstations/Dockerfile
```diff
@@ -0,0 +1,197 @@
...
+
+COPY --chmod=755 workstation-startup/* /etc/workstation-startup.d/
+COPY --chmod=644 jupyter/jupyter_notebook_config.json /opt/python/jupyter/etc/jupyter/jupyter_notebook_config.json
+
+### Clean up ###
+RUN apt-get remove -yq dpkg-sig \
+    && apt-get install -yqf --no-install-recommends \
+    && apt-get autoremove -yq \
+    && apt-get clean -yq \
+    && rm -rf /var/lib/apt/lists/*
```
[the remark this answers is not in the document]
ianpittwood: Yeah I'll grant that it doesn't do anything with the caches. It still cleans up unnecessary or broken packages that may otherwise land in the final container though so I'd like to leave it unless there's an advantage to copy-pasting it onto all the `apt` usages.

bschwedler (approved):

ianpittwood:
Setting as blocked by #583. The issue has been escalated to Google and we are awaiting their response.

workbench-for-google-cloud-workstations/.env
```diff
@@ -0,0 +1,12 @@
+RSW_VERSION=2023.03.2+454.pro2
+RSW_TAG_VERSION=2023.03.2-454.pro2
+RSW_DOWNLOAD_URL=https://download2.rstudio.org/server/bionic/amd64
+RSW_NAME=rstudio-workbench
+PYTHON_VERSION=3.10.12
+PYTHON_VERSION_ALT=3.9.17
+PYTHON_VERSION_JUPYTER=3.10.12
```
msarahan, line 7: This confused me, as I thought that it was a third installation of python. Why are we explicitly setting this? Why isn't it just always presumed to be installed at PYTHON_VERSION?

workbench-for-google-cloud-workstations/conf/repos.conf
```diff
@@ -0,0 +1,2 @@
+CRAN=https://packagemanager.posit.co/cran/__linux__/focal/latest
```
msarahan, line 1: It seems like we should be leaving this as the real CRAN URL, but making RSPM take priority. This is not especially a question about this PR, though, so perhaps it is best discussed elsewhere.

workbench-for-google-cloud-workstations/Dockerfile.ubuntu2004
```diff
@@ -0,0 +1,204 @@
...
+ENV DIAGNOSTIC_ENABLE false
+ENV DIAGNOSTIC_ONLY false
+ENV LICENSE_MANAGER_PATH /opt/rstudio-license
+ENV WORKBENCH_JUPYTER_PATH=/usr/local/bin/jupyter
+
+### Update/upgrade system packages ###
+RUN apt-get update --fix-missing  \
+    && apt-get upgrade -yq \
+    && apt-get install -yq --no-install-recommends \
+      apt-transport-https \
```
msarahan, line 33: can these be moved to an external file in the deps folder, like we do for R and python packages?

workbench-for-google-cloud-workstations/Dockerfile.ubuntu2004
```diff
@@ -0,0 +1,204 @@
...
+    && rm -f ./r-${R_VERSION}_1_amd64.deb \
+    && rm -f ./r-${R_VERSION_ALT}_1_amd64.deb \
+    && ln -s /opt/R/${R_VERSION}/bin/R /usr/local/bin/R \
+    && ln -s /opt/R/${R_VERSION}/bin/Rscript /usr/local/bin/Rscript \
+    && rm -rf /var/lib/apt/lists/*
+
+### Install Python versions ###
+RUN curl -O https://cdn.rstudio.com/python/ubuntu-2004/pkgs/python-${PYTHON_VERSION}_1_amd64.deb \
+    && curl -O https://cdn.rstudio.com/python/ubuntu-2004/pkgs/python-${PYTHON_VERSION_ALT}_1_amd64.deb \
+    && apt-get update \
```
msarahan, line 82: remove update since we did it above
```suggestion
```

workbench-for-google-cloud-workstations/Dockerfile.ubuntu2004
```diff
@@ -0,0 +1,204 @@
...
+      tcl \
+      tk \
+      tk-dev \
+      tk-table \
+    && rm -rf /var/lib/apt/lists/*
+
+### Install R versions ###
+RUN curl -O https://cdn.rstudio.com/r/ubuntu-2004/pkgs/r-${R_VERSION}_1_amd64.deb \
+    && curl -O https://cdn.rstudio.com/r/ubuntu-2004/pkgs/r-${R_VERSION_ALT}_1_amd64.deb \
+    && apt-get update \
```
msarahan, line 70: remove update since we did it above
```suggestion
```

workbench-for-google-cloud-workstations/Dockerfile.ubuntu2004
```diff
@@ -0,0 +1,204 @@
...
+      sssd \
+      supervisor \
+      tcl \
+      tk \
+      tk-dev \
+      tk-table \
+    && rm -rf /var/lib/apt/lists/*
+
+### Install R versions ###
+RUN curl -O https://cdn.rstudio.com/r/ubuntu-2004/pkgs/r-${R_VERSION}_1_amd64.deb \
```
msarahan, line 68: I wonder if it might be nicer to write a proper bash file here, and then use a loop for these duplicated lines. I'm ambivalent. It would be extra work. It would make the Dockerfile easier to read and update. It might make it easier to run/debug scripts outside of docker builds.

Up to you!

workbench-for-google-cloud-workstations/Dockerfile.ubuntu2004
```diff
@@ -0,0 +1,204 @@
...
+RUN mkdir -p /opt/rstudio-license/ \
+    && mkdir -p /var/lib/rstudio-workbench \
+    && curl -sL "https://s3.amazonaws.com/rstudio-ide-build/monitor/focal/rsp-monitor-workbench-gcpw-amd64-2023.06.0-419.pro1.tar.gz" |  \
+       tar xzvf - --strip 2 -C /opt/rstudio-license/ \
+    && chmod 0755 /opt/rstudio-license/license-manager \
+    && mv /opt/rstudio-license/license-manager /opt/rstudio-license/license-manager-orig \
+    && rm -f /usr/lib/rstudio-server/bin/license-manager
+
+### Install Jupyter and extensions ###
+RUN /opt/python/"${PYTHON_VERSION_JUPYTER}"/bin/python -m venv /opt/python/jupyter \
```
msarahan, line 154: A lot of this stuff has me feeling like we should have centralized scripts for this stuff. We must be duplicating a ton of this stuff between this and the other workstations image. I don't think it's important to consolidate it all here in this PR, but perhaps it is worth filing a tech debt ticket for.

msarahan (approved):
I think this is fine. I left some comments about things to talk about, but I don't think any immediate change is needed.

workbench-for-google-cloud-workstations/TurboActivate.dat
bschwedler, line 1: Is this file the same as is present in the other directories?
I know that we need to have this as a file, not a symlink, in the docker build context.  Would it make sense to have a single file that can moved into the correct location before building the container?

Definitely out of scope for this PR though.

workbench-for-google-cloud-workstations/conf/launcher-env
```diff
@@ -0,0 +1,6 @@
+JobType: session
+Environment: LANG=en_US.UTF-8
+    LANGUAGE=en_US:en
+    LC_ALL=en_US.UTF-8
+JobType: any
+Environment: PATH=/opt/python/3.9.16/bin:$PATH
```
bschwedler, line 6: Do we still need this? This caused some issues on the DataOps platform, especially because it was pre-pended to the path. 

If we do, we need to update this to the appropriate patch version.

workbench-for-google-cloud-workstations/docker-compose.test.yml
```diff
@@ -0,0 +1,24 @@
+version: '2.3'
+services:
+
+  sut:
+    image: $IMAGE_NAME
+    command: /run_tests.sh
+    entrypoint: []
+    environment:
+      # uses .env by default
```
bschwedler, line 9: ❤️

workbench-for-google-cloud-workstations/startup.sh
```diff
@@ -0,0 +1,68 @@
...
+
+# Activate License
+RSW_LICENSE_FILE_PATH=${RSW_LICENSE_FILE_PATH:-/etc/rstudio-server/license.lic}
+if [ -n "$RSP_LICENSE" ]; then
+    ${LICENSE_MANAGER_PATH}/license-manager activate $RSP_LICENSE || true
+elif [ -n "$RSP_LICENSE_SERVER" ]; then
+    ${LICENSE_MANAGER_PATH}/license-manager license-server $RSP_LICENSE_SERVER || true
+elif test -f "$RSW_LICENSE_FILE_PATH"; then
+    ${LICENSE_MANAGER_PATH}/license-manager activate-file $RSW_LICENSE_FILE_PATH || true
+fi
```
bschwedler, line 40: Can we either wrap this section with `set +x` so we don't end up logging the license key?

bschwedler (changes requested):

workbench-for-google-cloud-workstations/.env
ianpittwood: This matches the pattern of the other Workbench products. I personally have no problem defaulting to one version or the other, but we should probably make it a global change for the repo.

workbench-for-google-cloud-workstations/Dockerfile.ubuntu2004
ianpittwood: Yeah that would help slim down the Dockerfile itself, I'll do it. Perhaps we should apply this `deps/*` model to the other projects at some point.

workbench-for-google-cloud-workstations/Dockerfile.ubuntu2004
ianpittwood: I think we have to do it in each layer since we run `rm -rf /var/lib/apt/lists/*` to slim down at the end of each one.

workbench-for-google-cloud-workstations/TurboActivate.dat
ianpittwood: Though the files may seem identical, they are specific to each License Manager SKU. We can't share them across projects or things will a splode.

workbench-for-google-cloud-workstations/conf/launcher-env
ianpittwood: I have no clue! I noticed that these files tend to get wildly out of date in ever project that uses them. If we should keep it and update it, I'm happy to. If we can remove it, I would also be happy.

workbench-for-google-cloud-workstations/Dockerfile.ubuntu2004
ianpittwood: Same as above. I confirmed that we do need to do this each time if we remove the update lists at ever layer.

workbench-for-google-cloud-workstations/Dockerfile.ubuntu2004
ianpittwood: https://github.com/rstudio/rstudio-docker-products/issues/616

workbench-for-google-cloud-workstations/Dockerfile.ubuntu2004
ianpittwood: https://github.com/rstudio/rstudio-docker-products/issues/616

PR #566, opened 2023-06-27
Connect v2023.06.0 release

Lytol:
Upgrading Connect to 2023.06.0.  A few important notes:

- Running `just update-versions` is failing due to a missing file: `workbench-for-microsoft-azure-ml/Dockerfile.ubuntu1804`. It looks like the dockerfile was upgraded to Ubuntu 22 in <https://github.com/rstudio/rstudio-docker-products/commit/f2f306f367a12c11557bea33ae66f36ff790e76d>, but the Justfile was not updated accordingly.
- Additionally, it looks like the `R_VERSION` and `PYTHON_VERSION` were updated in the Justfile, but `update_versions` has not been run since... so other products will be unexpectedly updated if I do a full `just update-versions`

So, we used a targeted `just RSC_VERSION=2023.06.0 update-rsc-versions` to do the Connect version upgrade here.  I'm happy to do a PR for the issues above, but I don't fully know the repercussions for other teams, so I didn't want to make assumptions. Reach out to me if needed.


dbkegley:
I think the base branch should be `main` for this PR

Lytol:
Maybe that's where my issues are also coming from.  The default git branch is currently set to `dev`, so I was unclear whether I should be working out of `dev` or `main`...

Lytol:
I re-did this PR against `main` instead of `dev` and all the same issues exist. It's still unclear (from docs, etc) whether this should be branched off `main` or `dev`, so I'm leaving it alone until someone can clarify what should be done here.

ianpittwood (approved):
Correct, it should go to `dev` and then we'll merge `dev` to `main` to fully release once #463 is merged. This LGTM though. I'll go ahead and merge it in.

PR #571, opened 2023-06-29
Push "-dev" suffixed base builds for use by test and preview builds

### LINKED ISSUE #568: Push product base/pro builds for dev builds
#463 is currently failing tests because it's not able to use the rebuilt base images until they are in `main`. We should start pushing `dev` and other non-`main` branch builds to a `dev`/`preview` GHCR repository so we can use them in tests.

ianpittwood:
Closes #568 

Since moving our primary branch to `dev` we no longer get up to date base image builds that can be used by our CI. This change will allow our base builds to be pushed to `-dev` suffixed registries in GHCR for use in test and preview builds that may need newer base images that have not yet been merged into `main` to be built and pushed to the production registry.


ci.Justfile
```diff
@@ -47,13 +47,19 @@ build-base $OS $TYPE="base":
...
   elif [[ $TYPE == "base-pro" || $TYPE == "pro" || $TYPE == "product-base-pro" ]]; then
     IMAGE_NAME="product-base-pro"
+    SRC_IMAGE_NAME="product-base"
     CTX_PATH="./product/pro"
     FILE_PATH="./product/pro/Dockerfile.${OS}"
   fi
+  if [[ $BRANCH != "main" ]]; then
+    IMAGE_NAME="${IMAGE_NAME}-dev"
+    SRC_IMAGE_NAME="${SRC_IMAGE_NAME}-dev"
+  fi
```
bschwedler, line 62: Do we want to be pushing branch builds to a `-dev` tag, or should this be limited only to the `dev` branch?

ci.Justfile
```diff
@@ -92,16 +99,22 @@ build-base $OS $TYPE="base":
...
+test-base $OS $TYPE="base" $BRANCH=`git branch --show`:
   #!/usr/bin/env bash
   set -euxo pipefail
 
   # set short name
   if [[ $TYPE == "base" ]]; then
     IMAGE_NAME="product-base"
+    if [[ $BRANCH != "main" ]]; then
+      IMAGE_NAME="${IMAGE_NAME}-dev"
+    fi
```
bschwedler, line 111: Would it make more sense for us to move this logic to the top-level of the justfile, along with setting the `$BRANCH` variable for `just` commands?
Maybe something like `BRANCH_SUFFIX` so we don't have to have so many branching if conditions? Pun most definitely intended.

ci.Justfile
ianpittwood: The problem I foresaw is that any non-main/non-dev branch with a change to the base images may not be able to test if it can't push those up on build. We could consider adding logic for some alternate tags for non-main/non-dev branch builds so they could push up to GHCR and still be used in testing and then we could delete them afterwards.

ci.Justfile
ianpittwood: Normally, yes. But the whole of the file currently treats `$BRANCH` as a per-target argument. Modifying that behavior would require quite a few modifications to the workflows and possibly other inter-connecting Justfiles and docs. It's a good idea, but probably something that should become its own issue.

bschwedler (approved):

ci.Justfile
ianpittwood: After some thought, we'll change to only _pushing_ the images on `dev` and `main` builds. However, I'll keep this snippet since pretty much any branch that isn't `main` should likely use the `dev` branch builds.

ci.Justfile
```diff
@@ -116,11 +129,22 @@ build-release $PRODUCT $OS $VERSION $BRANCH=`git branch --show` $SHA_SHORT=`git
...
 
-  # set short name
+  # set short name and source image name
+  SRC_IMAGE_NAME=""
   if [[ $PRODUCT == "workbench" || $PRODUCT == "r-session-complete" || $PRODUCT == "workbench-for-microsoft-azure-ml" ]]; then
     SHORT_NAME="RSW"
+    if [[ $BRANCH == "main" ]]; then
+      SRC_IMAGE_NAME="product-base-pro"
+    else
+      SRC_IMAGE_NAME="product-base-pro-dev"
```
colearendt, line 139: It's worth noting that the conventions we currently have in place for this type of "preview" image are:
- name: `preview-<image-name>`
- tag: `<branch-name>-<tag>`

.github/workflows/build-latest.yaml
```diff
@@ -78,23 +78,15 @@ jobs:
           args: --file=./product/Dockerfile.${{ matrix.os }}
           command: test
 
-      - name: Login to Docker Hub
```
colearendt, line 81: No longer pushing to docker hub at all?

.github/workflows/build-preview.yaml
```diff
@@ -70,6 +70,13 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-buildx-
 
+      - name: Login to ghcr.io
```
colearendt, line 73: Why the order switch here?

colearendt (commented):
I realize it's already merged, but I'd love to understand some of the details here better 😄

.github/workflows/build-latest.yaml
ianpittwood: We never pushed to Docker Hub for base images. We could if there's value in doing so.

.github/workflows/build-preview.yaml
ianpittwood: We pull the base images from ghcr.io for most images now so we authenticate before starting the build. I could be wrong, but I think we need to be authenticated to pull or to pull without ratelimits.

PR #577, opened 2023-07-07
Add `cracklib-runtime` to base image package list

ianpittwood:
`cracklib-runtime` was briefly missing from the `ubuntu` Docker image, causing `chpasswd` calls to fail. It appears that this package has been added back into the `ubuntu` base image, but I don't think it will hurt to check for it and install it if it's missing in the future.

### DIFF
--- a/connect/test/goss.yaml
+++ b/connect/test/goss.yaml
@@ -12,6 +12,8 @@ group:
 package:
   rstudio-connect:
     installed: true
+  cracklib-runtime:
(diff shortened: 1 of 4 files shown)


aronatkins (approved):

ianpittwood:
@tdstein @aronatkins checked for us and verified `cracklib-runtime` was present in the latest images and that `chpasswd` works in those.

PR #579, opened 2023-07-07
Zack deprecate ubuntu18 connect

zackverham:


christierney (approved):

connect/NEWS.md
```diff
@@ -1,3 +1,7 @@
+# 2023-07-10
+
+- BREAKING: Deprecate the Ubuntu 18.04 (Bionic Beaver) images.
+
```
dbkegley, line 4: `connect-content-init` also has a NEWS.md file

colearendt (approved):
Looks awesome! Definitely recommend adding to the `connect-content-init` `NEWS.md` too though!

Also a small side effect that because this is separate from the Connect version bump, we will just stop producing a bionic build for a version, but that image will still exist (and be old).

connect/NEWS.md
zackverham: good catch - fixed!

PR #585, opened 2023-07-18
Move `license-manager-shim` copy step to after LM replacement

ianpittwood:
- Move `license-manager-shim` copy step to after the replacement of the default LM with Azure ML.
- Update Goss libssl1.0.0 sys package variable to libssl-dev for Ubuntu 22.04 upgrade
- Remove R 3.5.3 (archived) and add R 4.3.1
- Add `openssh-client` to Dockerfile system package dependencies


lagerratrobe (approved):
Looks good to me.

PR #588, opened 2023-07-20
Workflow refactor

### LINKED ISSUE #586: Refactor Github Action workflows for reusability and stability
### Background
Our workflows as they are currently designed have become a pain point for engineers. The current way we call `buildx` builds can cause CI failures on new images when they fail to load into the local Docker image cache for testing. Products are also very difficult to rebuild after initial release, regardless of whether they are the latest version or an older version with a vulnerability to patch. With some refactoring, we should be able to make our Workflows more reliable, stable, and easy to use.

### Goals
- [ ] Rebuild releases weekly
- [ ] Create a reusable Docker build/test/scan/push workflow that can be called by our other workflows
- [ ] Utilize official Github Actions where possible
- [ ] Ensure novel builds can still be tested and scanned prior to push
- [ ] Allow for products to more easily define different R/Python version combinations in image builds

ianpittwood:
Closes #586 

### Workflow Changes
- Creates
  - A new action, `.github/actions/build-test-scan-push`, which builds a specified image, tests it using Goss, scans it with Snyk, and pushes it. The last three steps are optional and can be enabled/disabled based on the needs of the user or workflow.
  - 3 new primary workflows
    - `build-release.yaml` - Builds the base image, base image w/ pro drivers, and then each product image utilizing the above action. This pipeline will trigger on merges and on a scheduled basis to keep images up to date with security patches, something we lacked on the previous revision. Replaces `build-latest.yaml`.
    - `build-prerelease.yaml` - Builds preview/daily images for each product as defined in the workflow matrix. Replaces `build-preview.yaml`.
    - `build-manual.yaml` - Allows repository administrators to perform one off manual builds of daily, preview, and release builds. This workflow is still in development and will likely need future work. The hope is that having this workflow will give us a quicker way of updating previously built images. Replaces `build-preview-webhook.yaml`.
  - Content image builds also now use `.github/actions/build-test-scan-push`, but have otherwise remained unchanged.
  - Adds retry logic for all builds to prevent one-off failures (e.g. network timeouts) from requiring human intervention.
  - Removed all `ci.Justfile` build shortcuts that were being used by previous workflows and replaces them with shortcuts for getting arguments and tags for each image.
- Miscellaneous Changes
  - Changes caching method from local caching to Github Actions caching. Github Actions caching should be less error prone for large images since it handles purges of older cache data for us.
  - Much of our usages of `just` have been removed in favor of official Github Action calls. The primary driver for this change was a bug with using `buildx` in Github Actions that caused properly exported images to disappear in later steps, causing things like tests to fail. 
  - Slight changes to workflow naming for readability.

### Other Changes
- Creates
  - A new target in the `Justfile` for updating Quarto versions across the project. This should stop Quarto from getting out of date across projects.
- Upgrades
  - All Python versions are now up to the latest security/bug fixes on their respective versions. The default versions have been changed to `3.9.17` and `3.8.17` (alternate). CentOS 7 builds are pinned back to `3.9.14` and `3.8.15`, the distribution's latest successful builds.
  - Defaults of R `3.6.x` have been removed and replaced with `4.2.3` and `4.1.3` (alternate).
- Miscellaneous Changes
  - Changed all usages of `JUPYTER_PYTHON_VERSION` to `PYTHON_VERSION_JUPYTER` for consistency.
  - Added `trimSuffix "-1"` for pro driver version Goss tests on RHEL-based systems.
  - Changed `rstudio-connect.gcfg` to replace Python versions with the primary version the image is built with at build time.
  - Cleans up version updates in Justfile and fixes blindspots. Ran after fixing so all versions are up to date.


Justfile
```diff
@@ -153,34 +158,42 @@ update-r-versions:
...
+    product/base/Dockerfile.ubuntu* \
+    product/pro/Dockerfile.ubuntu*
   sed {{ sed_vars }} "s/^R_VERSION := .*/R_VERSION := \"{{ R_VERSION }}\"/g" \
     workbench/Justfile \
     workbench-for-microsoft-azure-ml/Justfile \
-    connect/Justfile package-manager/Justfile \
+    connect/Justfile \
+    package-manager/Justfile \
+    product/base/Justfile \
+    product/pro/Justfile \
```
bschwedler, line 176: Would it be useful to grep for `^R_VERSION : =`  or `^R_VERSION=` to get the list of files to update instead of having to manually update them when new files are added?

We could even write a just command that would find the files matching a specific pattern and then send them as arguments to the sed command.

ci.Justfile
```diff
@@ -36,28 +38,39 @@ _get-default-tag PRODUCT OS:
 
   echo "{{ REGISTRY_NAMESPACE }}/${IMAGE_PREFIX}{{ PRODUCT }}:{{ OS }}"
 
-# just BUILDX_PATH=~/.buildx build-base ubuntu1804 base
-build-base $OS $TYPE="base" $BRANCH=`git branch --show`:
+# just _get-rsw-download-url release ubuntu1804
```
bschwedler, line 41:
```suggestion
# just _get-rsw-download-url release ubuntu2204
```

bschwedler (approved):
Thank you for doing this. Having common build steps across all the images should be very helpful.

I had just a few comments, but none of them need to be addressed before merging.

.github/workflows/build-prerelease.yaml
```diff
@@ -0,0 +1,108 @@
+on:
+  schedule:
+    # every morning at 8am UTC
+    # https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onschedule
+    # make sure to change the `if` commands below when changing the schedule
+    - cron: '0 8 * * *'
+    - cron: '0 9 * * *'
```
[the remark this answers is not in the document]
ianpittwood: Crons trigger off the "default branch" setup for the repo. For the pre-release builds we typically try to build one copy from `main` and one copy from `dev` (our default branch) so in order to do that we staggered the cron triggers and put an if in to choose the correct branch.

.github/actions/build-test-scan-push/action.yaml
```diff
@@ -0,0 +1,142 @@
...
+
+    - name: Test - ${{ inputs.test-image }}
+      shell: bash
+      run: |
+        if [[ "${{ inputs.test-image }}" == "true" ]]; then
+          echo "${{ inputs.build-args }}" > ${{ inputs.context }}/.env
+          echo "OS=${{ inputs.os }}" >> ${{ inputs.context }}/.env
+          cat ${{ inputs.context }}/.env
+          IMAGE_NAME=${{ steps.first-tag.outputs.FIRST_TAG }} docker-compose -f ${{ inputs.context }}/docker-compose.test.yml run sut
+        fi
```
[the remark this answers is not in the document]
ianpittwood: Yes, good idea. I'll put that in.

Justfile
ianpittwood: Yes I think we should spin off another ticket to clean these up.

ianpittwood: https://github.com/rstudio/rstudio-docker-products/issues/590

cm421 (approved):
LGTM! Thank you for refactoring all of this!

PR #608, opened 2023-08-07
Add `ca-certificates` and explicit `gpg` install to PPM build

ianpittwood:
GPG calls to keys.openpgp.org will fail without updated `ca-certificates`. We can explicitly install the certificates to prevent failures in signature verification.


kfeinauer:
Do we feel like `ca-certificates` belongs in the base image as well?

bschwedler (approved):
LGTM

PR #611, opened 2023-08-07
Remove ubuntu1804 base image

ianpittwood:
- Removes product/base/Dockerfile.ubuntu1804 and product/pro/Dockerfile.ubuntu1804 as neither are in use.
- Change defaults and example commands across all project Justfiles to ubuntu2204.
- Remove Dockerfile.*.draft from package-manager, would have to be rewritten if we decide to switch PPM to the repo product base image.
- Remove 18.04 builds from workflows.
- Add NEWS.md entry.


ianpittwood:
PPM build failures resolved by #612

product/pro/Justfile
```diff
@@ -3,7 +3,7 @@ set positional-arguments
 BUILDX_PATH := ""
 
 PRODUCT := "product-base-pro"
-IMAGE_OS := "ubuntu1804"
+IMAGE_OS := "ubuntu12204"
```
bschwedler, line 6:
```suggestion
IMAGE_OS := "ubuntu2204"
```

bschwedler (changes requested):

product/pro/Justfile
ianpittwood: ty

bschwedler (approved):

ianpittwood:
None of the products (besides PPM) still build Ubuntu 18.04 so this PR may be a bit misleading. 

All the Justfile changes are for example commands or arg defaults that still listed ubuntu1804 as the target OS. None of those Justfiles are actually used as a part of CI, they're just user/developer shortcuts when using the repo locally.

The product/base and product/pro images are the only ones that are actually having Ubuntu 18.04 removed in this PR and that's because nothing downstream is consuming them anymore so they're unnecessary and are causing us headaches trying to maintain them.

The Dockerfile.*.draft files removed from PPM were drafts that utilized the new base image to build PPM. We decided to stick with r-base for now though so they were just crowding the space. I decided to remove both for now since they would need to be rewritten anyways if we decided to go down that path in the future.

PR #612, opened 2023-08-07
Run `apt-get upgrade` on PPM image

ianpittwood:
The PPM image doesn't run `apt-get upgrade` at the moment which has caused problems with accessing certain sites due to out of date CA certs. This change adds an `apt-get upgrade` command and explicitly installs `ca-certificates` in case it isn't already present.


tylfin (approved):

bschwedler (approved):

package-manager/Dockerfile.ubuntu1804
```diff
@@ -10,16 +10,33 @@ ENV LANG en_US.UTF-8
...
+
 # Runtime settings ------------------------------------------------------------#
-ARG TINI_VERSION=0.18.0
-RUN curl -L -o /usr/local/bin/tini https://github.com/krallin/tini/releases/download/v${TINI_VERSION}/tini && \
-    chmod +x /usr/local/bin/tini
+ADD https://cdn.rstudio.com/platform/tini/v${TINI_VERSION}/tini-amd64 /tini
+ADD https://cdn.rstudio.com/platform/tini/v${TINI_VERSION}/tini-amd64.asc /tini.asc
+RUN gpg --update-trustdb \
+    && curl -sL -o /tmp/tini-key.asc "http://keyserver.ubuntu.com/pks/lookup?op=get&search=0x6380dc428747f6c393feaca59a84159d7001a4e5" \
+    && gpg --import /tmp/tini-key.asc \
```
bschwedler, line 33: Could you add a comment above the `RUN` command to explain why we are pulling the keys by fingerprint?

Until we remove the Ubuntu 18.04 images it would be good to know why this is different in this file.

ianpittwood: Yes, thank you for reminding me

PR #619, opened 2023-08-10
Expand default R and Python package lists for Workbench for Azure ML

### LINKED ISSUE #617: Update default package install list for Workbench for Azure ML
Azure ML has a set of Python packages to make integration of code with Azure ML resources easier. We should add these packages to our Workbench for Azure ML image so users may take advantage of them. We should also expand our default Python and R package sets to more broadly cover basic data science tools users in those instances may need.

Python packages:
```
azureml
azureml-core
azureml-dataset-runtime
azure-ai-ml
numpy
scipy
pandas
matplotlib
```

R packages:
```
DBI
R6
RJDBC
RODBC
RSQLite
Rcpp
base64enc
checkmate
crayon
curl
devtools
digest
ellipsis
evaluate
fastmap
ggplot2
glue
haven
highr
htmltools
htmlwidgets
httpuv
jsonlite
keyring
knitr
later
learnr
magrittr
markdown
mime
miniUI
mlflow
mongolite
odbc
openssl
packrat
plumber
png
profvis
promises
r2d3
ragg
rappdirs
rJava
readr
readxl
renv
reticulate
rgdal
rlang
rmarkdown
roxygen2
rprojroot
rsconnect
rstan
rstudioapi
sf
shiny
shinytest
sourcetools
stringi
stringr
testthat
tidyverse
tidymodels
tinytex
withr
xfun
xml2
xtable
yaml
```

ianpittwood:
Closes #617

- Adds Azure ML specific Python packages and more general data science packages for R and Python.
- Improves the organization of our dependencies for Workbench for Azure ML.


workbench-for-microsoft-azure-ml/Dockerfile.ubuntu2204
```diff
@@ -34,28 +34,16 @@ ENV DIAGNOSTIC_ONLY false
 ENV WORKBENCH_JUPYTER_PATH=/usr/local/bin/jupyter
 
 COPY --chmod=0644 TurboActivate.dat /opt/rstudio-license/license-manager.conf
+COPY deps/* /
```
bschwedler, line 37: Is there a better place to put these files than in the root directory? Perhaps even `/tmp`?

I know we clean them up in a later layer but this feels like the wrong place to put them.

workbench-for-microsoft-azure-ml/deps/install-r.sh
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -xe -o pipefail
 
+UBUNTU_CODENAME=$(lsb_release -cs)
```
bschwedler, line 4: Do we still need the `local` definition inside `install_r_packages`?

workbench-for-microsoft-azure-ml/deps/install-r.sh
```diff
@@ -38,7 +38,5 @@ for rvers in 3.6.3 4.0.5 4.1.3 4.2.3 4.3.1; do
     rm -f ./r-${rvers}_1_amd64.deb
 
     # install packages
-    install_r_packages /tmp/package-list.txt /opt/R/${rvers}/bin/R https://packagemanager.rstudio.com/cran/__linux__/bionic/latest
+    install_r_packages /tmp/r_packages.txt /opt/R/${rvers}/bin/R https://packagemanager.rstudio.com/cran/__linux__/${UBUNTU_CODENAME}/latest
```
bschwedler, line 41: Looks like we are already using `/tmp` here for other package lists.

workbench-for-microsoft-azure-ml/Dockerfile.ubuntu2204
ianpittwood: Yeah not sure why I made this change, I'll put it back in `/tmp`.

workbench-for-microsoft-azure-ml/deps/install-r.sh
ianpittwood: Good point, I'll double check and remove it if we don't.

bschwedler (approved):

PR #622, opened 2023-08-15
Push release pipeline on cron trigger

ianpittwood:
I thought that `push-image` would evaluate to true when ran on the cron trigger on `main`, but that was not the case. This should fix the problem by adding logic to evaluate to true when triggered by our cron schedule.


bschwedler (commented):
I wish we didn't have to specify the trigger schedule explicitly. 😞 

Can you please add a comment at the top of the workflow where the cron trigger is defined to remind future us to update the `push-image` logic?

bschwedler (approved):

PR #624, opened 2023-08-15
Remove unused Workbench Components

MariaSemple:
This PR removes unused Workbench components to reduce the surface area for security vulnerabilities in the images. It also adds an "RStudio Pro session"-only image for certain use-cases.

### DIFF
--- a/shared/remove-workbench-extras.sh
+++ b/shared/remove-workbench-extras.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+
+rm -rf /var/lib/rstudio-server/r-versions
+rm /usr/lib/rstudio-server/bin/rserver-saml
+rm /usr/lib/rstudio-server/bin/rserver-openid
+rm /usr/lib/rstudio-server/bin/librclient.so
+rm /usr/lib/rstudio-server/bin/librserver.so
+rm /usr/lib/rstudio-server/bin/license-manager
+rm /usr/lib/rstudio-server/bin/license-manager.conf
+rm /usr/lib/rstudio-server/bin/locktester
+rm /usr/lib/rstudio-server/bin/migrate.sh
+rm /usr/lib/rstudio-server/bin/pamtester 
+rm /usr/lib/rstudio-server/bin/rpcsend
+rm /usr/lib/rstudio-server/bin/rserver
+rm /usr/lib/rstudio-server/bin/rserver-acls
+rm /usr/lib/rstudio-server/bin/rserver-admin
+rm /usr/lib/rstudio-server/bin/rserver-db
+rm /usr/lib/rstudio-server/bin/rserver-encrypt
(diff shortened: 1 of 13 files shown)


shared/remove-workbench-extras.sh
```diff
@@ -0,0 +1,35 @@
...
+rm /usr/lib/rstudio-server/bin/rserver-monitor
+rm /usr/lib/rstudio-server/bin/rserver-pam
+rm /usr/lib/rstudio-server/bin/rserver-session-reaper
+rm /usr/lib/rstudio-server/bin/rstudio-kubernetes-launcher
+rm /usr/lib/rstudio-server/bin/rstudio-launcher
+rm /usr/lib/rstudio-server/bin/rstudio-launcher.sh
+rm /usr/lib/rstudio-server/bin/rstudio-local-launcher
+rm /usr/lib/rstudio-server/bin/rstudio-server
+rm /usr/lib/rstudio-server/bin/rstudio-slurm-launcher
+rm /usr/lib/rstudio-server/bin/run-diagnostics
```
ianpittwood, line 29: Should these also have `-f`?

workbench-session-rstudio-pro/remove-other-sessions.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+rm -rf /usr/lib/rstudio-server/bin/code-server
+rm -rf /usr/lib/rstudio-server/bin/vscode-workbench-ext
+rm /usr/lib/rstudio-server/bin/jupyter-session-run
+rm /usr/lib/rstudio-server/bin/vscode-session-run
```
ianpittwood, line 6: nit: no extra line

ianpittwood (changes requested):
Looks good so far. Just a few comments:
- We should update the `sed` targets in the root `Justfile` so versions get updated appropriately in the new `workbench-session-rstudio-pro` images.
- We should eventually add these builds into the workflows too. That should be as easy as adding in some more entries to the matrices.
- Is it possible we'll add more session images for Workbench? If so, it may be best to layer this in as a subdirectory like `workbench-session/rstudio-pro`.

shared/remove-workbench-extras.sh
MariaSemple: Happy to change it, but for some reason only the r-versions file had trouble

msarahan (approved):

shared/remove-workbench-extras.sh
ianpittwood: That's strange. Is it the only one present at the time this runs perhaps?

MariaSemple: No, I'm guessing is a permissions thing. The other ones are definitely present before this runs

PR #633, opened 2023-09-07
Add `STARTUP_DEBUG_MODE` flag for stack tracing

ianpittwood:
Turn off stack tracing by default


connect/startup.sh
```diff
@@ -1,7 +1,9 @@
 #!/bin/bash
 
 set -e
-set -x
+if [[ "${STARTUP_DEBUG_MODE}" -eq 1 ]]; then
```
bschwedler, line 4: Explicitly default this to OFF to prevent comparison problems.
```suggestion
if [[ "${STARTUP_DEBUG_MODE:-0}" -eq 1 ]]; then
```

bschwedler (approved):

PR #635, opened 2023-09-11
fix: update download path for PPM Dockerfiles

jmwoliver:
### Description

Updates the `Dockerfile.ubuntu2204` and `Dockerfile.ubuntu1804` PPM download paths to point to the correct URL. Also updates the `docker-compose.yml` to use `ubuntu2204` instead of `ubuntu1804`.

### DIFF
--- a/package-manager/Dockerfile.ubuntu1804
+++ b/package-manager/Dockerfile.ubuntu1804
@@ -49,7 +49,7 @@ RUN apt-get update -qq && \
 
 # Download RStudio Package Manager ---------------------------------------------#
 ARG RSPM_VERSION=2023.04.0-6
-ARG RSPM_DOWNLOAD_URL=https://cdn.rstudio.com/package-manager/ubuntu/amd64
+ARG RSPM_DOWNLOAD_URL=https://cdn.rstudio.com/package-manager/deb/amd64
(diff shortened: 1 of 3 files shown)


package-manager/Dockerfile.ubuntu1804
```diff
@@ -49,7 +49,7 @@ RUN apt-get update -qq && \
 
 # Download RStudio Package Manager ---------------------------------------------#
 ARG RSPM_VERSION=2023.04.0-6
-ARG RSPM_DOWNLOAD_URL=https://cdn.rstudio.com/package-manager/ubuntu/amd64
+ARG RSPM_DOWNLOAD_URL=https://cdn.rstudio.com/package-manager/deb/amd64
```
jmwoliver, line 52: I was going to delete this `Dockerfile.ubuntu1804` file, but I wasn't sure what all may use it. I just left it for now, let me know if this should end up being deleted.

tylfin: The arg version should be updated simultaneously as that version will no longer work with this install path

jmwoliver:
Closing in favor of https://github.com/rstudio/rstudio-docker-products/pull/637

PR #640, opened 2023-09-18
Pin jupyterlab version to most recent 3.6.x

melissa-barca:
As discussed in rstudio/rstudio-pro#3821, Workbench is not compatible with jupyter notebook 7.x and jupyterlab 4.x, which are now being installed by default with `pip`. Specifying the install version for Jupyterlab as 3.6.5, also ensures the latest version of notebook 6.x is installed.


cm421 (approved):
Assuming the different `3.2.9` jupyterlab version for centos7 is intentional, LGTM!

melissa-barca:
> Assuming the different `3.2.9` jupyterlab version for centos7 is intentional, LGTM!

yes, that is the most recent version available there

PR #643, opened 2023-09-27
fix main conflict

bschwedler:
- Bump up version for PPM to 2023.08.0-16
- Bump RSC_VERSION to 2023.09.0
- Update download URL too
- Bump up docker-compose.yml version and README.md version


ianpittwood (approved):

dbkegley:
@bschwedler Is this required because we are merging directly to `main` as part of the Connect release process?
https://github.com/rstudio/rstudio-docker-products/pull/638

We've recently updated our release scripts to create 2 separate PRs with basically the same contents. One of the PRs uses `main` as the base branch, and the other uses `dev` as the base, but they are created independently so they don't share any commits.

We do this so that we can bump the helm chart to the latest connect version which requires the images to be published to ghcr which I believe only happens after a push to main.

If there's a different process we should be following during our releases then please let us know and we can adapt our scripts

bschwedler:
@dbkegley There was a conflict in the merge to `main`, but it was for the workbench version so I don't see why it would have been caused by the two PRs for connect.

Given the challenges we've been having with the two-branch workflow, I think it would be beneficial to have a retro to figure out how we can make the process more efficient.
It feels like we've traded a set of challenges of bundling changes for another set of workflow challenges.

PR #653, opened 2023-11-01
Update PPM to 2023.08.4-20

glin:
I haven't done this in a while so just to be sure, I ran `just update-versions`. This updated the R versions as well. And am merging this to `dev` branch.

We skipped the last hotfix release, 2023.08.2, because that release didn't affect these Docker images. But 2023.08.4 contains some bug fixes that will affect users here.


bschwedler (approved):
LGTM

PR #658, opened 2023-11-08
Updates for Workbench 2022.07.4 backwards compatibility

ianpittwood:
* Add backwards compatibility for VS Code installation in older RSW versions
* Add option to force downloading from S3 where applicable
* Fix default tagging behavior in `workbench/Justfile`
* Use shared `_get-rsw-download-url` target in `workbench/Justfile`


kfeinauer (approved):
LGTM

PR #662, opened 2023-11-09
add Python to Package Manager images

shepherdjerred:
This PR adds Python and the `build` and `virtualenv` packages to Package Manager's Dockerfile. This is to support an upcoming feature.

* Don't build Package Manager for Ubuntu 18.04
* Update Package Manager's R from `4.2.0` to `4.2.3`
* Use `product-base` as the base image instead of `r-base`
  * This allows us to easily use Posit's Python builds and it aligns with what other the Dockerfiles are doing. Additionally, it allows us to simplify our Dockerfile quite a bit.
* Install `build` and `virtualenv`


jmwoliver (approved):

glin:
Should we use the Posit builds of Python? I'm not sure, but I think others will have the same question. The other images here all install Python from https://docs.posit.co/resources/install-python/ rather than the distro, e.g.:

https://github.com/rstudio/rstudio-docker-products/blob/5f47778a01441f00bdab433a0a603fe12704607a/product/base/Dockerfile.ubuntu2204#L115-L126

But just like with R, the Python version doesn't matter as much for PPM's git builders. Using Posit builds would just be more for consistency and in case users want to use Python for other things in the image.

package-manager/Dockerfile.ubuntu1804
```diff
@@ -24,8 +24,13 @@ RUN apt-get update --fix-missing \
       dirmngr \
       gpg \
       gpg-agent \
+      python3 \
```
colearendt, line 27: Should we be using system python? Or the python builds that we build ourselves? The latter would be more consistent with all of our other imgaes 😮

shepherdjerred:
I've updated the Package Manager Dockerfile to use the `product-base-pro` image to match what other products are doing.

We are still building Package Manager for Ubuntu 18.04. `product-base-pro` seems only to be built for Ubuntu 22.04. Do we still want to support 18.04? I know we've worked in the `package-manager` repository to drop support for 18.04.

package-manager/Dockerfile.ubuntu2204
```diff
@@ -1,70 +1,37 @@
...
+ARG REGISTRY=ghcr.io
+FROM ${REGISTRY}/rstudio/${SRC_IMAGE_NAME}:ubuntu2204-r${R_VERSION}_${R_VERSION_ALT}-py${PYTHON_VERSION}_${PYTHON_VERSION_ALT}
 LABEL maintainer="RStudio Docker <docker@rstudio.com>"
 
-ARG R_VERSION=4.2.3
+ARG PYTHON_VERSION=3.9.17
+ARG PYTHON_VERSION_ALT=3.8.17
 
 # Locale configuration --------------------------------------------------------#
-RUN localedef -i en_US -f UTF-8 en_US.UTF-8
```
shepherdjerred, line 8: All of these deleted lines are already in the base image.

package-manager/Dockerfile.ubuntu2204
```diff
@@ -1,70 +1,37 @@
...
-    && apt-get install -y --no-install-recommends ca-certificates gdebi-core gpg dpkg-sig \
-    && curl -O ${RSPM_DOWNLOAD_URL}/rstudio-pm_${RSPM_VERSION}_amd64.deb \
+RUN curl -O ${RSPM_DOWNLOAD_URL}/rstudio-pm_${RSPM_VERSION}_amd64.deb \
     # Pre 7/25/23 packages
     && gpg --keyserver keyserver.ubuntu.com --recv-keys 3F32EE77E331692F \
     # Post 7/25 packages
     && gpg --keyserver keys.openpgp.org --recv-keys 51C0B5BB19F92D60 \
     && dpkg-sig --verify rstudio-pm_${RSPM_VERSION}_amd64.deb \
-    && RSTUDIO_INSTALL_NO_LICENSE_INITIALIZATION=1 gdebi -n rstudio-pm_${RSPM_VERSION}_amd64.deb \
+    && RSTUDIO_INSTALL_NO_LICENSE_INITIALIZATION=1 apt-get install -y ./rstudio-pm_${RSPM_VERSION}_amd64.deb \
```
shepherdjerred, line 31: Use `apt-get` instead of `gdebi` so we don't have to install/uninstall `gdebi`

ci.Justfile
```diff
@@ -217,6 +223,12 @@ get-prerelease-args $TYPE $PRODUCT $OS $VERSION $BRANCH=`git branch --show`:
     else
       SRC_IMAGE_NAME="product-base-pro-dev"
     fi
+  elif [[ $PRODUCT == "package-manager" ]]; then
+    if [[ $BRANCH == "main" ]]; then
+      SRC_IMAGE_NAME="product-base"
+    else
+      SRC_IMAGE_NAME="product-base-dev"
+    fi
```
jmwoliver, line 231: What's the difference between `product-base` and `product-base-pro` (and their `-dev` variants)? Why does only Package Manager use `product-base` and the others use `product-base-pro` above?

jmwoliver (approved):

ci.Justfile
shepherdjerred: I'm not an expert here, but I believe that `product-base-pro` containers the Pro database drivers whereas `product-base` does not. Other products like Connect and Workbench have a use for these drivers, but Package Manager does not.

shepherdjerred:
@colearendt I'll wait for your sign-off on this since it's a fairly large change.

ianpittwood (approved):
Looks great! Thank you!

PR #672, opened 2023-12-05
avoid error with emtpy string (see Dockerfile)

adRn-s:
The env variable `RSW_TESTUSER_UID` was set to 10000 in previous versions, now it's an empty string by default, which makes this if statement to fail, I just saw it on my docker logs.

Swapping the order of the if statements makes it so that we only search for `RSW_TESTUSER_UID`'s username via getent in the case that we had the first switch (`RSW_TESTUSER`) set.


melissa-barca (changes requested):
Thank you for making this fix! Sorry for the delayed review. Could you bring the branch up to date?

adRn-s (commented):
I've just updated my branch with the latest commits from rstudio:dev

PR #692, opened 2024-01-30
Bump PWB to 2023.12.1+402.pro1

GCRev:


melissa-barca (approved):
LGTM

GCRev:
We can merge this but 2 (3?) of the checks failed.
One wasn't happy about `launcher.pem`'s read-write permissions
The other seemed like it ran out of drive space when building the image.

Do we want to consider those to be non-issues?

GCRev:
@ianpittwood could you assess whether we should merge based on the failures from the check runs?

ianpittwood:
@GCRev As you probably saw, I pushed a "fix" (removing R 3.6.3) for the Workbench AML definition so it stops erroring out. I need to do some testing locally for Workbench GCW to figure out why that's failing. Once that's done I think we should be good though. The daily build failure looked related to whatever Workbench build was used so I'm not concerned with that for now.

ianpittwood:
@GCRev We're good to go here. The GCW issue seems like a one-off.

GCRev:
Thank you!

PR #698, opened 2024-03-01
update quarto

kgartland-rstudio:
This PR updates to the latest quarto version used in Connect so the new jumpstarts can be deployed successfully.


aronatkins (commented):
Would you mind updating _all_ the Quarto version references?

There appear to be quite a few more, and not all associated with the `QUARTO_VERSION` variable.

```bash
git grep 1.3.450
```

Additionally, could you update at least the connect/NEWS.md and content/base/NEWS.md?

I don't know the effect of updating this version for other products; hopefully other folks can chime in.

PR #703, opened 2024-03-11
Use github-actions managed ubuntu-latest-4x runner for azureml builds

costrouc:
This PR allows building containers which exceed the 60GB storage max limit for github-actions. Eventually this PR should also allow parallel building of docker containers.  Potentially using docker buildx to allow docker to optimized the order in which layers are built.


ianpittwood:
@costrouc could we layer this on #701 and grant both AML and GCW builds the large builder? The two specialty images are the ones that struggle with space most frequently.

costrouc:
@ianpittwood I've layered this PR on the branch `fork-waml-ci-workflow`

ianpittwood (approved):

PR #705, opened 2024-03-12
Docker buildx bake experiment for builds

costrouc:
I have attempted to avoid any tricks or advanced features of hcl within bake. There are many optimizations we could make from here. Most importantly this does not break anything!

Docker buildx bake has several advantages:
 - allows us to represent the dependencies between dockerfiles
 - performs parallel builds (at the layer level)
 - allows us to cleanly express the build matricies that each product has
 - have a single maintainable place for managing all docker build configuration

This PR builds everything within a single runner and no complex github action workflows. All is captured within the `docker-bake.hcl`. It is possible to easily exclude certain builds etc and only build certain ones.

This builds:
    - product-base
    - product-base-pro
    - connect
    - connect-content-init
    - package-manager
    - r-session-complete
    - workbench
    - workbench-for-google-cloud-workstations
    - workbench-for-microsoft-azure-ml


costrouc:
The action that matters is `Build DockerImages with Bake`

atheriel:
This is interesting, I hadn't heard of `docker buildx bake`.

dbkegley:
FWIW we did a bunch of testing with `docker buildx bake` for Connect and found that the inline layer caching was unreliable for multi-stage builds. We think it was related to [this issue](https://github.com/moby/buildkit/issues/2274) which was just recently fixed. We have not re-tested yet though.

https://github.com/rstudio/connect/issues/23961#issuecomment-1613775719

ianpittwood:
> FWIW we did a bunch of testing with `docker buildx bake` for Connect and found that the inline layer caching was unreliable for multi-stage builds. We think it was related to [this issue](https://github.com/moby/buildkit/issues/2274) which was just recently fixed. We have not re-tested yet though.
> 
> [rstudio/connect#23961 (comment)](https://github.com/rstudio/connect/issues/23961#issuecomment-1613775719)

We weren't previously using caching for these builds so I'm not concerned with anything like that at the moment. Thanks for the heads up though!

ianpittwood:
Merged in #718

PR #707, opened 2024-03-14
Update Pro Drivers to 2024.03.0

bschwedler:
Use the latest version of the Professional Drivers


ianpittwood (approved):

ianpittwood (changes requested):
Is there a CentOS 7 release for this? I noticed those builds are failing for us.

bschwedler:
> Is there a CentOS 7 release for this? I noticed those builds are failing for us.
There is. The filename changed. I popped on a commit: `s/el7/el/`.

ianpittwood (approved):

PR #716, opened 2024-04-02
Extract common operations to reusable scripts in base images

ianpittwood:

PR #726, opened 2024-04-18
Dev to Main

colearendt:


ianpittwood (changes requested):
@colearendt before we do this can we finish off these?
* https://github.com/rstudio/rstudio-docker-products/pull/716
* I'll create a follow-up very shortly to upgrade Connect in our bake definitions and then make the Quarto adjustments requested by @dbkegley and the Connect team.

Once those two are reviewed and merged, I think we're good to merge up.

ianpittwood (approved):

PR #727, opened 2024-04-18
Move Quarto installation/configuration to downstream images

ianpittwood:
Previously we installed Quarto in the base image which then carried to each downstream product image. As a result, it was very difficult to have products with diverging Quarto versions. After #716 though we now have shared scripts baked into the base image to perform common tasks, such as installing a specific Quarto version. Connect needs to diverge from the Quarto versions used by other images so this change finalizes moving Quarto installations from the base image to downstream product images (should they need to or choose to install Quarto).

- Upgrade Connect's Quarto version to 1.4.552
- Update Workbench images to add their bundled version of Quarto to the `PATH`
- Pin remaining products back to 1.3.340 or lower where necessary
- Make testing around Quarto install and symlinking more robust
- Do not install Quarto in PPM


docker-bake.hcl
```diff
@@ -97,7 +97,7 @@ variable PACKAGE_MANAGER_BUILD_MATRIX {
 variable CONNECT_BUILD_MATRIX {
     default = {
         builds = [
-            {os = "ubuntu2204", r_primary = "4.2.3", r_alternate = "4.1.3", py_primary = "3.9.17", py_alternate = "3.8.17"},
+            {os = "ubuntu2204", r_primary = "4.2.3", r_alternate = "4.1.3", py_primary = "3.9.17", py_alternate = "3.8.17", quarto = "1.4.552"},
```
kfeinauer, line 100: Does the quarto version again have to be redefined here or can it share the environment definition?

dbkegley:
This is great, it's going to be so much easier for us to manage these software version changes now. Thanks for all your work on this @ianpittwood! 

Since we are between Connect releases, can we wait on the quarto upgrade since it would be a breaking change for those who are using the current release? If we bump the quarto version in the `2023.03` image then customers who are already using the latest version will need to update their config file otherwise Connect would fail to start. We can make this a clean upgrade starting with `2024.04` and we'll add the quarto upgrade as a breaking change in the docker image's news for the next release. Now that the quarto installation is decoupled from the base images I can manage that upgrade during the next release

ianpittwood:
> This is great, it's going to be so much easier for us to manage these software version changes now. Thanks for all your work on this @ianpittwood!
> 
> Since we are between Connect releases, can we wait on the quarto upgrade since it would be a breaking change for those who are using the current release? If we bump the quarto version in the `2023.03` image then customers who are already using the latest version will need to update their config file otherwise Connect would fail to start. We can make this a clean upgrade starting with `2024.04` and we'll add the quarto upgrade as a breaking change in the docker image's news for the next release. Now that the quarto installation is decoupled from the base images I can manage that upgrade during the next release

OK. I'll modify this PR to keep `1.3.340` for Connect. Then we can still merge it and then merge dev to main.

dbkegley (approved):

PR #733, opened 2024-04-22
Fix failing bake push for Workbench for Microsoft Azure ML

ianpittwood:


costrouc:
Why is the build push failing currently? It looks like you've removed a few targets right?

ianpittwood:
I removed the `waml-images` group in favor of calling `workbench-for-microsoft-azure-ml`. Since `workbench-for-microsoft-azure-ml` inherits the other builds, they don't need to be explicitly included as a group for the build. `workbench-for-microsoft-azure-ml` is the only target in that build line that has tags, gets exported, and pushed so it also fixes the problem where the bake GHA can't push the other artifacts because they aren't tagged.

costrouc (approved):
Awesome thanks for explaining why this PR was needed!

PR #740, opened 2024-04-30
Build and bake previews every hour

tylfin:
With manual builds running into issues, we'd like more frequent checks of our latest dev/RC builds to avoid delays.


kfeinauer (approved):
We will most likely revert this when the manual builds are fixed. They are currently breaking.

https://github.com/rstudio/rstudio-docker-products/actions/runs/8883145104

Tagging @ianpittwood

PR #764, opened 2024-05-08
Bump r version to 4.4 to address CVE and alt version to 4.3.3

GCRev:
Hopefully these are all the correct occurrences. Ian recommended that we also bump the alt version to 4.3.3.


docker-bake.hcl
```diff
@@ -96,10 +96,10 @@ function get_tags {
...
     default = {
         builds = [
-            {os = "centos7", r_primary = "4.2.3", r_alternate = "4.1.3", py_primary = "3.9.14", py_alternate = "3.8.15"},
-            {os = "ubuntu2204", r_primary = "4.2.3", r_alternate = "4.1.3", py_primary = "3.9.14", py_alternate = "3.8.15"},
-            {os = "ubuntu2204", r_primary = "4.2.3", r_alternate = "4.1.3", py_primary = "3.9.17", py_alternate = "3.8.17"},
-            {os = "ubuntu2204", r_primary = "4.2.3", r_alternate = "4.1.3", py_primary = "3.12.1", py_alternate = "3.11.7"},
+            {os = "centos7", r_primary = "4.4.0", r_alternate = "4.3.3", py_primary = "3.9.14", py_alternate = "3.8.15"},
+            {os = "ubuntu2204", r_primary = "4.4.0", r_alternate = "4.3.3", py_primary = "3.9.14", py_alternate = "3.8.15"},
+            {os = "ubuntu2204", r_primary = "4.4.0", r_alternate = "4.3.3", py_primary = "3.9.17", py_alternate = "3.8.17"},
+            {os = "ubuntu2204", r_primary = "4.4.0", r_alternate = "4.3.3", py_primary = "3.12.1", py_alternate = "3.11.7"},
```
ianpittwood, line 102: Only add one new build with the updated R and Python versions, please. Updating older builds will have unintended consequences for other products.

docker-bake.hcl
```diff
@@ -111,15 +111,15 @@ variable PRO_BUILD_MATRIX {
 variable PACKAGE_MANAGER_BUILD_MATRIX {
     default = {
         builds = [
-            {os = "ubuntu2204", r_primary = "4.2.3", r_alternate = "4.1.3", py_primary = "3.12.1", py_alternate = "3.11.7"},
+            {os = "ubuntu2204", r_primary = "4.4.0", r_alternate = "4.3.3", py_primary = "3.12.1", py_alternate = "3.11.7"},
```
ianpittwood, line 114: I imagine PPM would be fine with this, but let's keep this PR focused on Workbench.

docker-bake.hcl
```diff
@@ -111,15 +111,15 @@ variable PRO_BUILD_MATRIX {
...
         ]
     }
 }
 
 variable CONNECT_BUILD_MATRIX {
     default = {
         builds = [
-            {os = "ubuntu2204", r_primary = "4.2.3", r_alternate = "4.1.3", py_primary = "3.9.17", py_alternate = "3.8.17", quarto = DEFAULT_QUARTO_VERSION},
+            {os = "ubuntu2204", r_primary = "4.4.0", r_alternate = "4.3.3", py_primary = "3.9.17", py_alternate = "3.8.17", quarto = DEFAULT_QUARTO_VERSION},
```
ianpittwood, line 122: Please remove this. Connect should not have dependencies modified outside of a release.

docker-bake.preview.hcl
```diff
@@ -114,8 +114,8 @@ function get_tags {
 variable BASE_BUILD_MATRIX {
     default = {
         builds = [
-            {os = "centos7", r_primary = "4.2.3", r_alternate = "4.1.3", py_primary = "3.11.9", py_alternate = "3.10.14"},
-            {os = "ubuntu2204", r_primary = "4.2.3", r_alternate = "4.1.3", py_primary = "3.11.9", py_alternate = "3.10.14"},
+            {os = "centos7", r_primary = "4.4.0", r_alternate = "4.3.3", py_primary = "3.11.9", py_alternate = "3.10.14"},
+            {os = "ubuntu2204", r_primary = "4.4.0", r_alternate = "4.3.3", py_primary = "3.11.9", py_alternate = "3.10.14"},
```
ianpittwood, line 118: This matrix is fine. The previews can be kept on the latest version since they're primarily for testing and internal use.

docker-bake.preview.hcl
```diff
@@ -127,15 +127,15 @@ variable PRO_BUILD_MATRIX {
...
 }
 
 variable CONNECT_BUILD_MATRIX {
     default = {
         builds = [
-            {os = "ubuntu2204", r_primary = "4.2.3", r_alternate = "4.1.3", py_primary = "3.11.9", py_alternate = "3.10.14", quarto = DEFAULT_QUARTO_VERSION},
+            {os = "ubuntu2204", r_primary = "4.4.0", r_alternate = "4.3.3", py_primary = "3.11.9", py_alternate = "3.10.14", quarto = DEFAULT_QUARTO_VERSION},
         ]
     }
 }
```
ianpittwood, line 141: Also should be fine to update these products exclusively for previews.

r-session-complete/Dockerfile.ubuntu2204
```diff
@@ -1,8 +1,8 @@
 FROM product-base-pro as build
 
 ARG DEBIAN_FRONTEND=noninteractive
-ARG R_VERSION=4.2.3
-ARG R_VERSION_ALT=4.1.3
+ARG R_VERSION=4.4.0
+ARG R_VERSION_ALT=4.3.3
```
ianpittwood, line 5: These variables could also get updated in the Workbench images, but I'll eventually remove the defaults because these get overwritten by `bake`.

ianpittwood (changes requested):
We'll need to remove some of the changes in `docker-bake.hcl` to keep this isolated to Workbench. Other products will be handled separately.

docker-bake.hcl
GCRev: Sorry. I wasn't sure how this system works, so I just did a blanket replace.

r-session-complete/Dockerfile.ubuntu2204
GCRev: Is changing the variables here necessary if the bake always overwrites them? Or should we keep this in?

ianpittwood: It's not necessary, but the change isn't a problem either so I wouldn't ask you to go roll them back.

docker-bake.hcl
ianpittwood, line 1: It looks like we lost the changes for Workbench in here, unless we're doing that separately when the 2024.04.1 patch is released.

docker-bake.preview.hcl
```diff
@@ -115,7 +115,9 @@ variable BASE_BUILD_MATRIX {
     default = {
         builds = [
             {os = "centos7", r_primary = "4.2.3", r_alternate = "4.1.3", py_primary = "3.11.9", py_alternate = "3.10.14"},
```
ianpittwood, line 117: Remove this entry, please. We can make all the previews use the same dependency versions for their respective OSes.

docker-bake.preview.hcl
```diff
@@ -115,7 +115,9 @@ variable BASE_BUILD_MATRIX {
...
+            {os = "centos7", r_primary = "4.4.0", r_alternate = "4.3.3", py_primary = "3.11.9", py_alternate = "3.10.14"},
             {os = "ubuntu2204", r_primary = "4.2.3", r_alternate = "4.1.3", py_primary = "3.11.9", py_alternate = "3.10.14"},
```
ianpittwood, line 119: Same here, remove this and make all the downstream build matrices use the same R/Py dependency versions.

ianpittwood (changes requested):

ianpittwood (approved):

cm421 (approved):
LGTM!

PR #765, opened 2024-05-09
Set the bakefile dynamically for manual builds

ianpittwood:


.github/workflows/build-manual.yaml
```diff
@@ -95,15 +104,48 @@ jobs:
...
+          fi
+          
+          product="${{ inputs.product }}"
+          if [[ "$product" == "connect" ]] || [[ "$product" == "connect-content-init" ]] || [[ "$product" == "content-images" ]]; then
+            product="CONNECT"
+          elif [[ "$product" == "package-manager" ]]; then
+            product="PACKAGE_MANAGER"
+          else
+            product="WORKBENCH"
+          fi
```
bschwedler, line 145: Will there ever be a case where `inputs.product` is unset?

bschwedler (approved):

.github/workflows/build-manual.yaml
ianpittwood: No, it's dropdown with `required: true` so that should be impossible.

PR #780, opened 2024-05-28
Enable user group creation for `rstudio-user` in Workbench image

ianpittwood:


cm421 (approved):
LGTM!

workbench/startup.sh
```diff
@@ -66,7 +66,10 @@ else
     if [ -z "$RSW_TESTUSER" ]; then
         echo "Empty 'RSW_TESTUSER' variables, not creating test user";
     else
-        useradd -m -s /bin/bash -N -u $RSW_TESTUSER_UID $RSW_TESTUSER
+        if [ -z "$RSW_TESTUSER_UID" ]; then
+            RSW_TESTUSER_UID=1000
```
kfeinauer, line 70: Did you say this used to default to 10,000? I'm wondering if we want to keep that - seems a potentially better chance for less collision?

ianpittwood: Fixed

kfeinauer (approved):

PR #799, opened 2024-06-19
minor fix to prevent if clause from erroring if 'getent passwd' returns a string with spaces

michaelmayer2:
The current check `if [ $(getent passwd $RSW_TESTUSER_UID) ]` in line 63 of startup.sh errors out if `RSW_TESTUSER_UID` exists and the returning string contains some spaces, e.g. 

```
root@83e35ab0ef5d:/# getent passwd ubuntu
ubuntu:x:1000:1000:AWS ParallelCluster user:/home/ubuntu:/bin/bash
root@83e35ab0ef5d:/# if [ $(getent passwd ubuntu) ]; then echo "fail"; fi
bash: [: ParallelCluster: binary operator expected
root@83e35ab0ef5d:/# if [ "$(getent passwd ubuntu)" ]; then echo "fail"; fi
fail
```

This very minor patch prevents this from happening again.

### DIFF
--- a/workbench/startup.sh
+++ b/workbench/startup.sh
@@ -60,7 +60,7 @@ unset RSW_LICENSE
 unset RSW_LICENSE_SERVER
 
 # Create one user
-if [ $(getent passwd $RSW_TESTUSER_UID) ] ; then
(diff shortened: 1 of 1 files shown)


ianpittwood (approved):
LGTM

PR #813, opened 2024-07-10
Workbench configuration and certificate bug patches

ianpittwood:
- Since Workbench's debian package currently attempts to start Workbench as a service during builds, a race condition can be created where Workbench begins to initialize and cannot finish before the builder kills the process during stage cleanup. This can cause files to be created with incorrect permissions. This change adds a 30sec delay after Workbench's installation for it to finish initializing. It also adds an immediate successor step to remove any certificate files that may have been generated.
- `launcher-env` files are fixed to point to `/opt/python/default` for the Python path.
- Adds `launcher.pem` and `launcher.pub` files to the `.dockerignore` and `.gitignore` files to ensure they are not accidentally added to image builds.


kfeinauer (approved):

cm421 (approved):
LGTM

PR #814, opened 2024-07-15
Manage TinyTeX installation with Quarto

ianpittwood:
Add CLI options to `install_quarto.sh` to install, update, add to path, or uninstall TinyTeX

Add an option to use Workbench's preinstalled Quarto for TinyTeX management


product/base/scripts/ubuntu/install_quarto.sh
```diff
@@ -72,4 +104,33 @@ install_quarto() {
...
+    if [[ "$ADD_PATH_TINYTEX" -eq 1 ]]; then
+        $QUARTO_PATH install tinytex --update-path --no-prompt
+    else
+        $QUARTO_PATH install tinytex --no-prompt
+    fi
+}
+
+if [[ "$QUARTO_PATH" != "/lib/rstudio-server/bin/quarto/bin/quarto" ]]; then
+    # Skip installation if Quarto is bundled with Workbench
+    install_quarto
```
melissa-barca, line 126: We might be able to remove some logic around here because it's safe to assumed quarto will always be bundled with Workbench. Historically it wasn't supported in rhel/centos 7, but quarto added support and now workbench has dropped support.

ianpittwood: I changed the script to more of an "auto-detect" approach rather than a flag. If `/lib/rstudio-server/bin/quarto/bin/quarto` exists, then we'll assume we should use that instead of installing another Quarto version.

I still need some logic in there regardless since these scripts are utilized by the Connect image.

product/base/Dockerfile.ubuntu2204
```diff
@@ -30,14 +30,6 @@ RUN gpg --batch --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 595E85A6B1
...
     && ln -s /tini /usr/local/bin/tini
 
-### Install TinyTeX ###
-SHELL ["/bin/bash", "-o", "pipefail", "-c"]
-RUN curl -sL "https://yihui.org/tinytex/install-bin-unix.sh" | sh \
-    && /root/.TinyTeX/bin/*/tlmgr path remove \
-    && mv /root/.TinyTeX/ /opt/TinyTeX \
-    && /opt/TinyTeX/bin/*/tlmgr option sys_bin /usr/local/bin \
-    && /opt/TinyTeX/bin/*/tlmgr path add
-
```
bschwedler, line 40: Thank you for removing this method!

product/base/scripts/ubuntu/install_quarto.sh
```diff
@@ -16,18 +16,34 @@ usage() {
...
+IS_WORKBENCH_INSTALLATION=0
 
-OPTIONS=$(getopt -o hdr: --long help,debug,prefix: -- "$@")
+# Set Quarto Path to the bundled version in Workbench if it exists
+if [ -f "/lib/rstudio-server/bin/quarto/bin/quarto" ]; then
+    QUARTO_PATH="/lib/rstudio-server/bin/quarto/bin/quarto"
+    IS_WORKBENCH_INSTALLATION=1
+fi
+
+OPTIONS=$(getopt -o hdr: --long help,debug,prefix:,use-workbench-quarto,install-tinytex,add-path-tinytex,update-tinytex,uninstall-tinytex -- "$@")
```
bschwedler, line 46: Can we remove parsing of the options that we aren't handling?
```suggestion
OPTIONS=$(getopt -o hd --long help,debug,prefix:,install-tinytex,add-path-tinytex,update-tinytex,uninstall-tinytex -- "$@")
```

bschwedler (approved):
LGTM

Feel free to merge after updating the shell option handling.

product/base/scripts/ubuntu/install_quarto.sh
ianpittwood: Good catch!

bschwedler (approved):

PR #831, opened 2024-08-29
Use `--fail` flag for curl to prevent silent failures

ianpittwood:
- Add common `-fsSL` flags to curl usage throughout project.
- Fix Quarto tests for changes from 3-part to 2-part versions.

PR #833, opened 2024-09-04
Fix Workbench and Connect builds

melissa-barca:
The Workbench and Connect builds have been failing: https://github.com/rstudio/rstudio-docker-products/actions/workflows/build-bake-preview.yaml

This is due to the regex that confirms tinytex is installed failing. Currently, this is expecting a 3 part release number, but tinytex has switched to two digit numbers: https://github.com/rstudio/tinytex-releases/releases

I kept the check for the third piece as optional to cover future releases.

### DIFF
--- a/connect/test/goss.yaml
+++ b/connect/test/goss.yaml
@@ -120,4 +120,4 @@ command:
     title: quarto_tinytex_installed
     exit-status: 0
     stderr:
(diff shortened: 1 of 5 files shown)


melissa-barca:
Changes already exist in https://github.com/rstudio/rstudio-docker-products/pull/831

PR #836, opened 2024-09-06
Replaced broken sed command (on macs) with ed alternative.

### LINKED ISSUE #790: Updating RSC_VERSION fails on MacOS
During the latest release of Connect we found that our release automation break when run on MacOS due to this line:

https://github.com/rstudio/rstudio-docker-products/blob/c635200ffddbda7b8eb6d2ec4027ac58872de0df/Justfile#L277

```
just RSC_VERSION=2024.05.1 update-rsc-versions
+ sed -i '' 's/RSC_VERSION=.*/RSC_VERSION=2024.05.1/g' connect/.env connect/Dockerfile.ubuntu2204 connect-content-init/Dockerfile.ubuntu2204
+ sed -i '' 's/^RSC_VERSION := .*/RSC_VERSION := "2024.05.1"/g' Justfile
++ just _get-clean-version 2024.05.1
+ sed -i '' -E 's/[0-9]{4}\.[0-9]{1,2}\.[0-9]{1,2}/2024.05.1/g' connect/README.md connect-content-init/README.md
+ sed -i '/variable CONNECT_VERSION/!b;n;c\ \ \ \ default = "2024.05.1"' docker-bake.hcl
sed: 1: "docker-bake.hcl": extra characters at the end of d command
error: Recipe `update-rsc-versions` failed with exit code 1
```

We tried replacing the `-i` with `{{ sed_vars }}` but that doesn't seem to fix the issue either. Something to do with the new `!b;n;c` isn't playing nicely with BSD sed on mac. Right now we can only run this automation on Linux but most of our developers responsible for releases are on mac.

plascaray:
Fixes #790

Fixed the broken `sed` command with an `ed` alternative. Spent some time trying to get an alternative `sed` command to work but I was not able to find one.

The new `ed` command finds the `CONNECT_VERSION` line, moves to the next line and changes the whole line with the new content, then stops processing changes, saves and quits.

I noticed that all blocks for each product use the same pattern, I can update the rest when the changes are reviewed / approved.

I also noticed that this notation seems to break syntax highlighting on VSCode. I have an alternative `awk` implementation that does not have that issue (it requires a file swap so I favored `ed` since it does the change in place):

```
  awk -v new_version="{{ RSC_VERSION }}" '
  /variable CONNECT_VERSION/ { print; getline; print "    default = \"" new_version "\""; next }
  { print }
  ' docker-bake.hcl > file.tmp && mv file.tmp docker-bake.hcl
```


bschwedler (approved):
Thank you so much for tackling this!

I wouldn't worry about syntax highlighting breaking in VS Code. That seems to happen any time I have to do anything somewhat complex in a `justfile`.

As long as the `ed` or `awk` command works on both Mac and Linux, I am happy with using either solution for all of the updates.

If you could please implement this for the other update-version commands, that would be most excellent!
Thanks!

dbkegley:
@bschwedler since Connect is planning a release for later this week, would it be possible to get these changes merged into `main` as well?

bschwedler:
Most definitely!

PR #845, opened 2024-09-24
Bump Workbench to v2024.09.0

zachhannum:

PR #846, opened 2024-09-24
Update R & Python for Workbench

bschwedler:
Relates to #845 
- **Update to latest R & Python patch versions for Workbench**


bschwedler:
@ianpittwood Do you see any issues with this update?

docker-bake.hcl
```diff
@@ -99,6 +99,7 @@ variable BASE_BUILD_MATRIX {
             {os = "ubuntu2204", r_primary = "4.2.3", r_alternate = "4.1.3", py_primary = "3.9.17", py_alternate = "3.8.17"},
             {os = "ubuntu2204", r_primary = "4.2.3", r_alternate = "4.1.3", py_primary = "3.12.1", py_alternate = "3.11.7"},
```
ianpittwood, line 100: I think these can be removed, but otherwise this seems good.

ianpittwood (changes requested):

ianpittwood (approved):

PR #847, opened 2024-09-24
Stop writing launcher debug logs to container

bschwedler:
Additional debug logging can be enabled through logging.conf if required


bschwedler:
@zachhannum Do you have any concerns of including this change with the Workbench release?

zachhannum (approved):
Nope, LGTM!

PR #861, opened 2024-10-25
Always install and upgrade pip and setuptools on WAML image

ianpittwood:


bschwedler (approved):
Thank you for digging in to fix that pesky ❌ !

PR #862, opened 2024-10-29
Add Workbench session init container image

skyeturriff:
Part of https://github.com/rstudio/rstudio-pro/issues/6920

Hoping to get some early(ish) feedback on adding a new image. I know this might be a bit awkward given this repo is in a bit of transition, so want to help mitigate any headaches/future pains where I can and make sure I'm doing things properly!

This image will be the new init container as part of Workbench's efforts to improve session components. I used `connect-content-init` as my example for how to set this up.

So far I've only added a new bake target for a daily build. I added this to the `build-manual` and `build-bake-preview` github workflows.

**NOTE** This also updates the `rstudio_workbench_daily` branch to kousa dogwood in `get-version.py`

I've tested this locally via

```bash
just preview-bake workbench-session-init-daily
mkdir init
docker run --rm -v $(pwd)/init:/mnt/init rstudio/workbench-session-init-preview:workbench-session-init-jammy-2024.11.0-daily-328.pro3
```

I also included a basic goss test.

### DIFF
--- a/.github/workflows/build-bake-preview.yaml
+++ b/.github/workflows/build-bake-preview.yaml
@@ -421,3 +421,43 @@ jobs:
           dockerhub-token: ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
           snyk-org: ${{ secrets.SNYK_ORG }}
           snyk-token: '${{ secrets.SNYK_TOKEN }}'
+
+  workbench-session-init-daily:
+    needs: [versions]
+    name: Workbench Session Init - Daily
+    runs-on: ubuntu-latest-4x
+
+    env:
+      target: "workbench-session-init-daily"
+      WORKBENCH_DAILY_VERSION: ${{ needs.versions.outputs.WORKBENCH_DAILY_VERSION }}
+      WORKBENCH_PREVIEW_VERSION: ${{ needs.versions.outputs.WORKBENCH_PREVIEW_VERSION }}
+      PACKAGE_MANAGER_DAILY_VERSION: ${{ needs.versions.outputs.PACKAGE_MANAGER_DAILY_VERSION }}
+      PACKAGE_MANAGER_PREVIEW_VERSION: ${{ needs.versions.outputs.PACKAGE_MANAGER_PREVIEW_VERSION }}
+      CONNECT_DAILY_VERSION: ${{ needs.versions.outputs.CONNECT_DAILY_VERSION }}
+      BRANCH: ${{ github.head_ref || github.ref_name }}
+
+    concurrency:
+      group: bake-workbench-session-init-daily-${{ github.ref }}
+      cancel-in-progress: true
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+        id: setup-buildx
+        with:
+          buildkitd-config: ./share/buildkitd.toml
+
+      - name: Build, Test, and Push
+        uses: ./.github/actions/bake-test-push
+        with:
+          target: ${{ env.target }}
+          bakefile: docker-bake.preview.hcl
+          push-image: ${{ github.ref == 'refs/heads/main' || github.ref == 'refs/heads/dev' || github.ref == 'refs/heads/dev-rspm' }}
+          ghcr-token: ${{ secrets.GITHUB_TOKEN }}
+          dockerhub-username: ${{ secrets.DOCKER_HUB_USERNAME }}
+          dockerhub-token: ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
+          snyk-org: ${{ secrets.SNYK_ORG }}
+          snyk-token: '${{ secrets.SNYK_TOKEN }}'
--- a/workbench-session-init/Dockerfile.ubuntu2204
+++ b/workbench-session-init/Dockerfile.ubuntu2204
@@ -1,0 +1,26 @@
+FROM ubuntu:noble AS build
+
+# Install required tools:
+# - ca-certificates installs necessary certificates to use cURL with HTTPS websites
+# - curl is used to download the runtime tar.gz
+ARG DEBIAN_FRONTEND=noninteractive
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends ca-certificates curl && \
+    rm -rf /var/lib/apt/lists/*
+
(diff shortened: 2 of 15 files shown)


workbench-session-init/Dockerfile.ubuntu2204
```diff
@@ -0,0 +1,25 @@
...
+    apt-get install -y --no-install-recommends ca-certificates curl && \
+    rm -rf /var/lib/apt/lists/*
+
+ARG RSW_VERSION=2024.09.0+375.pro3
+
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+RUN mkdir -p /pwb-staging && \
+    RSW_VERSION_URL=$(echo -n "${RSW_VERSION}" | sed 's/+/-/g') && \
+    echo "Downloading https://s3.amazonaws.com/rstudio-ide-build/session/multi/x86_64/rsp-session-multi-${RSW_VERSION_URL}-x86_64.tar.gz" && \
+    curl -fsSL -o /pwb-staging/rsp-session-multi.tar.gz "https://s3.amazonaws.com/rstudio-ide-build/session/multi/x86_64/rsp-session-multi-${RSW_VERSION_URL}-x86_64.tar.gz" && \
```
zachhannum, line 17: I actually tweaked the naming of this a little bit before merging my PR.

```suggestion
    curl -fsSL -o /pwb-staging/rsp-session-multi.tar.gz "https://s3.amazonaws.com/rstudio-ide-build/session/multi/x86_64/rsp-session-multi-linux-${RSW_VERSION_URL}-x86_64.tar.gz" && \
```

zachhannum (approved):
Should get approval from someone on platform, but I think it would be good to go ahead and get this into dailies for QA and testing! IMO we can push up additional PRs to add the release build and swap out the run script for something more elegant.

bschwedler:
Will this also help us reduce what is included in the session image?
- https://github.com/rstudio/rstudio-docker-products/pull/624

.github/workflows/build-bake-preview.yaml
```diff
@@ -421,3 +421,43 @@ jobs:
...
+    name: Workbench Session Init - Daily
+    runs-on: ubuntu-latest-4x
+
+    env:
+      target: "workbench-session-init-daily"
+      WORKBENCH_DAILY_VERSION: ${{ needs.versions.outputs.WORKBENCH_DAILY_VERSION }}
+      WORKBENCH_PREVIEW_VERSION: ${{ needs.versions.outputs.WORKBENCH_PREVIEW_VERSION }}
+      PACKAGE_MANAGER_DAILY_VERSION: ${{ needs.versions.outputs.PACKAGE_MANAGER_DAILY_VERSION }}
+      PACKAGE_MANAGER_PREVIEW_VERSION: ${{ needs.versions.outputs.PACKAGE_MANAGER_PREVIEW_VERSION }}
+      CONNECT_DAILY_VERSION: ${{ needs.versions.outputs.CONNECT_DAILY_VERSION }}
```
bschwedler, line 436: Do we need to specify the version for products we aren't building in the workflow?

I don't yet fully understand if all bake targets require all these env vars.

workbench-session-init/test/goss.yaml
```diff
@@ -0,0 +1,13 @@
+package:
+  curl:
+    installed: true
+
+file:
+  /opt/session-components/:
+    exists: true
+    mode: "0755"
+    filetype: directory
```
bschwedler, line 9: If there are specific session component files, we should probably check for them here.

If we want to check the indivdual contents of a tarfile that was added to the container, we could use a [command](https://goss.readthedocs.io/en/stable/gossfile/#command) block, running `tar` with the `-t` flag to ensure that the tarball has the expected contents.

workbench-session-init/run.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+set -e
```
bschwedler, line 3: If we can, I'd love for this to more strongly error out when there is a failure in the script.
```suggestion
set -euo pipefail
```

bschwedler (approved):
This looks good overall. Thanks for adding these!

.github/workflows/build-bake-preview.yaml
bschwedler: @ianpittwood This is really a question for you.

skyeturriff:
> Will this also help us reduce what is included in the session image?

@bschwedler That's the goal, yes! We'll use the init container to load all the session-specific components( vscode, rsession, rserver-launcher, etc.) into the session image, so the session image itself only needs to include the basics (system dependencies, R, python, etc.). Zach's prototype includes an example "base" session image [here](https://github.com/rstudio/rstudio-pro/blob/main/.devcontainer/containers/session-core/Dockerfile.rhel8)

.github/workflows/build-bake-preview.yaml
ianpittwood: Yeah. It's a bake limitation. It doesn't handle null values on build variables even if they're not part of the specified targets.

workbench-session-init/test/goss.yaml
skyeturriff: Ooh, I think the issue here is the tarball gets cleaned up in the last line of the Dockerfile `RUN` command, so it won't be around for goss to test against. But I think it's good practice to verify this, so I could check against the extracted directory structure. I'm uncertain how this would behave across versions though, if say we removed or added contents to the archive in a future release. I think in that case, we'd need to update the test, which would then only run successfully against the latest version?

ianpittwood (approved):
LGTM

skyeturriff:
I'm going to merge this in as-is so we can start to get dailies for testing, and address what's remaining (adding the release build, swapping out the run script, a more robust goss test) in follow-up PR(s)

zachhannum:
@ianpittwood @bschwedler does the repo need to be created manually on the dockerhub side of things? https://github.com/rstudio/rstudio-docker-products/actions/runs/11611641762/job/32333438273#step:4:1398

ianpittwood:
> @ianpittwood @bschwedler does the repo need to be created manually on the dockerhub side of things? https://github.com/rstudio/rstudio-docker-products/actions/runs/11611641762/job/32333438273#step:4:1398

Oh yes, let me go do that for you guys.

ianpittwood:
@zachhannum @skyeturriff it had a successful build now to Dockerhub.

PR #863, opened 2024-11-01
Add release target for workbench-session-init image

skyeturriff:
Follow-up to https://github.com/rstudio/rstudio-docker-products/pull/862 to add the release build for the Workbench session init container image.

We're only building the session components for 2024.11.0, so I added a new `WORKBENCH_SESSION_INIT_VERSION` variable that gets the daily version. Once we release 2024.11.0 and update `RSW_VERSION` in this repo, we can remove it and update the target steps to use the default release version.

I think this will also require a new repo `rstudio/workbench-session-init` to be added to our Docker hub.

Part of https://github.com/rstudio/rstudio-pro/issues/6920


zachhannum (approved):
LGTM!

bschwedler (approved):

skyeturriff:
I'm wondering if it would be better to hold off and put this in  along with the updates we normally do for the Workbench release. That way we wouldn't need the interim `WORKBENCH_SESSION_INIT_VERSION` variable as a workaround for the fact that the release version is currently still 2024.09

PR #864, opened 2024-11-04
Add positron config files and fix vs code default setting

melissa-barca:
Add positron files: 
- positron.conf
- positron.extensions.conf
- positron-user-settings.json

to r-session-complete, workbench, workbench for google cloud, and workbench for azure images.
Also replaced a VS Code setting that has been deprecated with the new setting.


cm421 (approved):
LGTM!

PR #875, opened 2024-11-20
Merge versions job into workbench-session-init job

ianpittwood:
Since `versions` only handles the version for `workbench-session-init`, it makes sense to have them build together.


skyeturriff (approved):
lgtm! Thank you!

PR #876, opened 2024-11-26
Workbench session image promotion

bschwedler:
Build the `workbench-session` image.


docker-bake.hcl
```diff
@@ -209,6 +225,8 @@ group "default" {
         "package-manager",
         "r-session-complete",
         "workbench",
+        "workbench-session",
+        "workbench-session-init",
```
zachhannum, line 229:
```suggestion
```

Can we just remove it from here to prevent it from building?

bschwedler: #878 Resolves this at the Actions level.

zachhannum (approved):

bschwedler:
@zachhannum is it OK if the preview/daily builds still run for the session init container? (as they did in the above workflow)

zachhannum:
> @zachhannum is it OK if the preview/daily builds still run for the session init container? (as they did in the above workflow)

Yep, preview/daily builds are still OK! We are using them for QA prior to the release. Thanks!

zachhannum:
> @zachhannum is it OK if the preview/daily builds still run for the session init container? (as they did in the above workflow)

Yep, preview/daily builds are still OK! We are using them for QA prior to the release. Thanks!

PR #878, opened 2024-11-27
Update build-bake.yaml

skyeturriff:
Removal of release `workbench-session-init` job from GitHub actions for the merge to `main`.

### DIFF
--- a/.github/workflows/build-bake.yaml
+++ b/.github/workflows/build-bake.yaml
@@ -333,6 +333,7 @@ jobs:
           snyk-token: '${{ secrets.SNYK_TOKEN }}'
 
   workbench-session-init:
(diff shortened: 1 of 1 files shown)


zachhannum (changes requested):
Instead of deleting the entire job, I would just add:

```{yaml}
if: false
```

To prevent the job from running under all conditions. Then we can remove that line when we want to enable it again.

https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idif

skyeturriff:
Ah, thank you! I should have done some research first 🥹 I can make the changes this aft

bschwedler:
@skyeturriff I amended the commit and can shepherd this through the PRs today.

PR #880, opened 2024-12-16
Enable workbench-session-init release target

skyeturriff:
Addresses https://github.com/rstudio/rstudio-pro/issues/7038

This should be merged in after/at the same time as `RSW_VERSION` is updated to 2024.12.

Removes the `WORKBENCH_SESSION_INIT_VERSION` var that was being used as a temporary workaround to grab the daily version, replacing it with `WORKBENCH_VERSION`. Also re-enables the `workbench-session-init` release target.


melissa-barca (approved):
LGTM!

zachhannum (approved):
LGTM!

PR #890, opened 2025-01-24
Upgrade actions versions

ianpittwood:
Fixes bake-action incompatibility with newer buildx versions


jmwoliver (approved):
Thanks for this! Our dogfood sites aren't getting the latest images right now because this GHA is borked.

Issue #891, opened 2025-01-28
Add `gdb` to session images


kevinushey:
Mainly for use at https://dev.palm.ptd.posit.it, so we can debug session crashes, collect core dumps, and the like.

(We can use the RStudio Terminal to attach gdb to the currently running session, and then have that produce a core dump when a deadly signal is sent -- and then follow up with analysis afterwards.)

kevinushey:
Bread crumbs:

https://github.com/rstudio/rstudio-docker-products/blob/2237dad6c77981de42cff924fbc72312083685b3/docker-bake.preview.hcl#L419

which I think implies this is the Dockerfile being used?

https://github.com/rstudio/rstudio-docker-products/blob/dev/workbench/Dockerfile.ubuntu2204

Issue #887, opened 2025-01-15
Add `strace` to Workbench session images


fh-mthomson:
To enable customers who leverage Posit's Workbench images to be able to more readily turn on `strace` diagnostics. 

Example: https://docs.posit.co/ide/server-pro/vscode_sessions/troubleshooting.html

PR #892, opened 2025-01-28
add gdb, strace to workbench Dockerfile

kevinushey:
Closes https://github.com/rstudio/rstudio-docker-products/issues/891.
Maybe also https://github.com/rstudio/rstudio-docker-products/issues/887?

I'm also not sure if I'm targeting the correct Dockerfiles 😅 The goal is to have the session images at https://dev.palm.ptd.posit.it updated.

### DIFF
--- a/workbench/Dockerfile.ubuntu2204
+++ b/workbench/Dockerfile.ubuntu2204
@@ -29,6 +29,7 @@ ENV WORKBENCH_JUPYTER_PATH=/usr/local/bin/jupyter
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 RUN apt-get update \
     && apt-get install --no-install-recommends -y \
+      gdb \
       krb5-user \
       libcap2 \
       libcurl4-gnutls-dev \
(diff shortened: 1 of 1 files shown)


cm421:
This would technically work, though I suspect a more appropriate location would be [in the base image package list](https://github.com/rstudio/rstudio-docker-products/blob/dev/product/base/deps/ubuntu2204_packages.txt). That would extend `gdb` and `strace` to all of the product images.

That's a very unilateral decision, however, which would likely start a lengthy back-and-forth discussion across all the teams. We can put it here for now, but given there's another issue making this request, we should at least start the conversation in the meantime

cm421 (approved):
Approving for now. If other teams are open to including it we can move it into the base or pro images later

fh-mthomson:
thanks for adding this! 

For context on #887 - I'd asked Posit Support (#111763):
> We're building our Docker images off of Posit's:
Server: https://github.com/rstudio/rstudio-docker-products/tree/main/workbench 
Session: https://github.com/rstudio/rstudio-docker-products/tree/main/r-session-complete
Which of these does strace need to be added to? Both?

The recommendation was to add to `r-session-complete`, rather than `workbench`. That was, indeed, minimally sufficient for us to then generate `strace` logs  in Workbench :)

PR #902, opened 2025-02-25
Use latest url when acquiring workbench daily versions

zachhannum:
Our preview/daily images are all still on 2024.12, so we don't have any 2025.04 preview images for testing. I _think_ this is what we need to update to accomplish this?

I could update the URL to `/mariposa-orchid/`, but unless anyone has any objections, I can't see a reason not to just use `latest` (although I also fully appreciate that we are hoping to abandon this repo sooner rather than later).

Testing this locally gives me:

```
./tools/get-version.py workbench --type daily --local 
Swapping product 'workbench' for 'workbench'
Providing version for product: 'workbench' and version type: 'daily'
Got daily version: '2025.04.0-daily+328.pro2'
2025.04.0-daily+328.pro2
```


zachhannum:
Looks like there were some build failures on the Workbench side for version 2025.04.0-daily+328.pro2 which are the source of the Workbench Session Init check failing, unrelated to the changes in this PR.

cm421:
I suspect things could get weird when we're close to releasing and are actively building the release branch and the `main` branch. I don't remember right off how the dailies page handles "daily" (aka "latest" in this context) builds when it's actively building the new release branch and `main` at the same time.

cm421 (approved):

zachhannum:
> I suspect things could get weird when we're close to releasing and are actively building the release branch and the `main` branch. I don't remember right off how the dailies page handles "daily" (aka "latest" in this context) builds when it's actively building the new release branch and `main` at the same time.

I think latest is always whichever branch we are promoting as "featured": https://github.com/rstudio/latest-builds/blob/5cfb6c51ce6dc9bbd0fcc0cce7b5dfdaedc4ab32/layouts/latest/list.json.json#L4

That seems reasonable to me; don't we generally update that to point to whichever release we are currently working on?

PR #906, opened 2025-03-11
Fix TinyTeX permissions on Workbench images

ianpittwood:
This fixes a potential bug where calls to tools like `pandoc` or `rmarkdown` by non-root processes where a TinyTeX package is called will throw a permissions error. This is similar to the behavior described in #904. `quarto install tinytex` always installs to `$HOME/.TinyTeX` with no way to change the install path via the Quarto CLI. This means that our current install process, which runs as `root`, will install TinyTeX to `/root/.TinyTeX` making TinyTeX and its packages inaccessible to non-root users. An issue for modifying the behavior of Quarto's TinyTeX install path already exists in https://github.com/quarto-dev/quarto-cli/issues/11800.

Changes:
- Prefix Quarto installation with `HOME="/opt"` to trick Quarto into install TinyTeX to `/opt/.TinyTeX`. This will make TinyTeX and its packages globally read/execute accessible.
- Update Goss tests to check
  - `perl` is installed as it is required for `tlmgr` to work.
  - TinyTeX install directory exists and is in `/opt`.
  - Existence/permissions on a few expected TinyTeX binaries. `pdflatex` and its links are specifically checked as it was the one reported for this bug.
  - Symlinks exist as expected in `/usr/local/bin` for TinyTeX.
  - Quarto identifies TinyTeX as an "External Installation".
- Added note in `NEWS.md` about the change, though I do not expect it to change the behavior of currently functional content deployments.

### DIFF
--- a/r-session-complete/Dockerfile.ubuntu2204
+++ b/r-session-complete/Dockerfile.ubuntu2204
@@ -22,6 +22,7 @@ RUN apt-get update \
       libuser \
       libuser1-dev \
       libpq-dev \
+      perl \
       rrdtool \
       strace \
       subversion \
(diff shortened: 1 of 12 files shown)


bschwedler (approved):

cm421 (approved):
Thanks for taking this on! LGTM!

PR #908, opened 2025-03-19
Remove bin/node from pwb init deps

zachhannum:
As of https://github.com/rstudio/rstudio-pro/commit/9171cb058328bf5c8cdfae073661e252101ffb31, node is no longer included as a standalone dep in our session components. This PR removes it from the dependency list to prevent copy failures.

I also added to the goss tests to explicitly check for each file/directory dependency we have a hard check for in the Go code so that hopefully we can catch removed dependencies earlier in the future.


jeffvroom (approved):
LGTM!

PR #916, opened 2025-04-21
Less chowning.

ssinnott:
Ok idea here is that we don't chown the files that are transferred over for adhoc jobs. This fixes an permission issue that we were seeing when the init container starts because the init container in this case is not run as root.


zachhannum (approved):
LGTM!

PR #919, opened 2025-05-06
Release Workbench 2025.05.0+496.pro5

adamhigerd:
I'm going to need to know if there's any action items I need to take after this gets merged.


cm421 (approved):
Linter conniptions aside, LGTM!

We should eventually follow the linter's suggestions, as they mostly look like sane suggestions. But I don't think that's your responsibility today

bschwedler (approved):

PR #923, opened 2025-05-28
Don't special-case workbench-session-init in the manual build

zachhannum:
This PR removes the special case for manually building workbench-session-init, since it produces an environment variable `WORKBENCH_SESSION_INIT_<TYPE>_VERSION` which is not used by any of the bake targets, which actually just expect `WORKBENCH_<TYPE>_VERSION` like the other workbench-related image targets.

## Broken Build

https://github.com/rstudio/rstudio-docker-products/actions/runs/15305976326

## Fixed Build (run from this PR branch)

https://github.com/rstudio/rstudio-docker-products/actions/runs/15308025160

### DIFF
--- a/.github/workflows/build-manual.yaml
+++ b/.github/workflows/build-manual.yaml
@@ -141,8 +141,6 @@ jobs:
             product="CONNECT"
           elif [[ "$product" == "package-manager" ]]; then
             product="PACKAGE_MANAGER"
-          elif [[ "$product" == "workbench-session-init" ]]; then
-            product="WORKBENCH_SESSION_INIT"
           else
             product="WORKBENCH"
           fi


ianpittwood:
> The Workbench Session Init Daily build is failing in this PR because of unrelated upstream build failures.
> 
> I'd like to get this PR to produce builds for our patch release, which should be fine, so ideally we can get this merged in even with the failing PR check.

@zachhannum is this issue documented somewhere? Are we not constructing that URL correctly?

zachhannum:
> > The Workbench Session Init Daily build is failing in this PR because of unrelated upstream build failures.
> > I'd like to get this PR to produce builds for our patch release, which should be fine, so ideally we can get this merged in even with the failing PR check.
> 
> @zachhannum is this issue documented somewhere? Are we not constructing that URL correctly?

It's because we determine the version to build based on the latest _Workbench_ build: https://github.com/rstudio/rstudio-docker-products/blob/21fddb789c2a33693ef07542cd17fe12e86bfd1c/tools/get-version.py#L101

But even if the jammy Workbench build succeeds, there are downstream builds in our Jenkins pipelines that can fail and cause that version to not be published for the various products, e.g. the session components. The Workbench session init container is particularly susceptible because it relies on _all_ of the session components platforms to succeed.

I don't believe we have issue (in this repo or otherwise) to track this.

We should probably set the daily version for the workbench-session-init container separate from the main Workbench version. Do you want me to try and address that as a part of this PR?

ianpittwood:
@zachhannum ah yes, that's much more clear! I would like the issue to get addressed, but I don't have a preference on adding to this PR or creating a new one. Just let me know if I should review this now or wait for you to make additional modifications.

astayleraz (approved):
LGTM

PR #932, opened 2025-06-06
Connect quarto version

markrtucker:


ianpittwood:
@dbkegley is this dangerous to do mid-release?

dbkegley:
Yes, sadly. The Connect helm chart also depends the specific version of quarto: https://github.com/rstudio/helm/blob/72ff22a35496fa46780c48084be3457b8fd4b721/charts/rstudio-connect/values.yaml#L420-L425

Bumping the quarto/r/python version mid-release would break the helm chart because the default configs wouldn't align: https://github.com/rstudio/helm/blob/72ff22a35496fa46780c48084be3457b8fd4b721/Justfile#L89-L118

We typically try to release interpreter version updates at the same time as a product release to avoid migration issues for customers

bschwedler:
@markrtucker is this still necessary? If so, can we apply it to the latest product version and images?

markrtucker:
> @markrtucker is this still necessary? If so, can we apply it to the latest product version and images?

The Chronicle team would still like this. For context, PTD uses these docker images and the Chronicle Cookbook requires Quarto 1.5.x-ish. This means that we cannot publish the Cookbook with source code to Connect in PTD.

I don't fully understand the concerns noted above by @dbkegley, so I don't know how to adequately address those concerns.

Ultimately, we would like to see the connect docker image include a recent version of quarto (1.4 is about 18 months old at this point).

dbkegley:
> > @markrtucker is this still necessary? If so, can we apply it to the latest product version and images?
> 
> The Chronicle team would still like this. For context, PTD uses these docker images and the Chronicle Cookbook requires Quarto 1.5.x-ish. This means that we cannot publish the Cookbook with source code to Connect in PTD.
> 
> I don't fully understand the concerns noted above by @dbkegley, so I don't know how to adequately address those concerns.
> 
> Ultimately, we would like to see the connect docker image include a recent version of quarto (1.4 is about 18 months old at this point).

The quarto version defined in the Connect image is used by customers who run Connect with local execution. If we bump the quarto version in the official connect image then it will break customer installations with existing configurations referencing quarto v1.4.557. This is why we typically bump the default r/python/quarto installations in conjunction with a Connect release because it means that we won't push a breaking change to an existing image tag.

Doesn't PTD use off-host execution? If so then could you add a new execution environment using one of the images available under the `content-base` repo? One of those images should have a compatible quarto version. The latest has quarto 1.7

https://github.com/rstudio/rstudio-docker-products/pkgs/container/content-base

stevenolen:
I don't see any problem with this. Is this configurable via the connect UI, or in config files?

bschwedler (changes requested):
These changes are stale.

PR #933, opened 2025-06-06
Fix out of date PPM versions

ianpittwood:
- Fix the `update*versions` targets in the `Justfile` since many referenced files that have been removed.
- Run `just update-rspm-versions` to update missed locations like `docker-bake.hcl`.


ianpittwood:
@CDRayn this is a follow-up to #930. The `docker-bake.hcl` file was missed in the version updates so the CI is still held back at `2025.04.0`. This will fix that issue and get the images pushed up at the current release.

CDRayn:
@ianpittwood Ah, my bad for missing that. Thanks for submitting the fix!

CDRayn (approved):

PR #936, opened 2025-06-18
Enhancements to license file documentation and link fixes

### LINKED ISSUE #734: Update License File Instructions
We should move to depreciate the `license-manager activate-file` command to what has been recommended by the product teams in the admin guide.

License files should be the easiest activation method for containers because as long as the file ends up in the correct format with the correct permissions under /var/lib/<rstudio-product> it should license without issue and shouldn't have the risk of bleeding activations due to bad container exits. 

In the future, when we move all customers to license files, we could think up a way to have customers mount the license files directly into `/var/lib/<rstudio-product>` and remove the activation steps from the startup.sh because the deactivation trap isn't needed for license files.

## Connect 
https://docs.posit.co/connect/admin/licensing/#license-file-activation

https://github.com/rstudio/rstudio-docker-products/blob/4bef13545b9ccee9748ef0aad22b8f7cbe6ab476/connect/startup.sh#L39
```bash
chown rstudio-server $RSC_LICENSE_FILE_PATH
chmod 600 $RSC_LICENSE_FILE_PATH
cp -a $RSC_LICENSE_FILE_PATH /var/lib/rstudio-server/.
```

## PPM
Commands not published until next release: 
https://github.com/rstudio/package-manager/blob/58070369ab1d2a8bf460dac6952d9c7bb3378824/docs/main/docs/admin/licensing.md?plain=1#L84-L99

https://github.com/rstudio/rstudio-docker-products/blob/4bef13545b9ccee9748ef0aad22b8f7cbe6ab476/package-manager/startup.sh#L40
```bash
chown rstudio-pm $RSPM_LICENSE_FILE_PATH
chmod 0600 $RSPM_LICENSE_FILE_PATH
cp -a $RSPM_LICENSE_FILE_PATH /var/lib/rstudio-pm/
```
I don't think we need to do anything special with the `--userspace` flag, but we should verify with PPM

## Workbench
https://docs.posit.co/ide/server-pro/license_management/license_activation_and_details.html#activation  

https://github.com/rstudio/rstudio-docker-products/blob/4bef13545b9ccee9748ef0aad22b8f7cbe6ab476/workbench/startup.sh#L53
```bash
chown root $RSW_LICENSE_FILE_PATH
chmod 0600 $RSW_LICENSE_FILE_PATH
cp -a $RSW_LICENSE_FILE_PATH /var/lib/rstudio-connect/
```

**

ianpittwood:
Closes #734 

- Expand documentation of using license files for product activation in containers.
- Fixes for various broken links.
- Most remaining mentions of "RStudio" replaced with "Posit".
- General refinement and improvement of READMEs for consistency and readability on Docker Hub.

### DIFF
--- a/connect/startup.sh
+++ b/connect/startup.sh
@@ -36,12 +36,11 @@ if ! [ -z "$RSC_LICENSE" ]; then
 elif ! [ -z "$RSC_LICENSE_SERVER" ]; then
     /opt/rstudio-connect/bin/license-manager license-server $RSC_LICENSE_SERVER
     trap deactivate EXIT
+elif test -f "$RSC_LICENSE_FILE_PATH"; then
+    rm -f /var/lib/rstudio-connect/*.lic || true
+    /opt/rstudio-connect/bin/license-manager activate-file $RSC_LICENSE_FILE_PATH
 elif ls /var/lib/rstudio-connect/*.lic >/dev/null 2>&1; then
     echo "Detected a license file in /var/lib/rstudio-connect/*.lic."
-elif test -f "$RSC_LICENSE_FILE_PATH"; then
-    rm -f /var/lib/rstudio-connect/*.lic
-    cp "${RSC_LICENSE_FILE_PATH}" /var/lib/rstudio-connect/license.lic
-    chmod g-rwx,g-rwx /var/lib/rstudio-connect/license.lic
 fi
 
 # ensure these cannot be inherited by child processes
--- a/connect/startup.sh
+++ b/connect/startup.sh
@@ -36,6 +36,8 @@ if ! [ -z "$RSC_LICENSE" ]; then
 elif ! [ -z "$RSC_LICENSE_SERVER" ]; then
     /opt/rstudio-connect/bin/license-manager license-server $RSC_LICENSE_SERVER
     trap deactivate EXIT
+elif ls /var/lib/rstudio-connect/*.lic >/dev/null 2>&1; then
+    echo "Detected a license file in /var/lib/rstudio-connect/*.lic."
 elif test -f "$RSC_LICENSE_FILE_PATH"; then
     rm -f /var/lib/rstudio-connect/*.lic
     cp "${RSC_LICENSE_FILE_PATH}" /var/lib/rstudio-connect/license.lic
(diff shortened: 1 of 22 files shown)


connect/startup.sh
```diff
@@ -36,6 +36,8 @@ if ! [ -z "$RSC_LICENSE" ]; then
 elif ! [ -z "$RSC_LICENSE_SERVER" ]; then
     /opt/rstudio-connect/bin/license-manager license-server $RSC_LICENSE_SERVER
     trap deactivate EXIT
+elif ls /var/lib/rstudio-connect/*.lic >/dev/null 2>&1; then
+    echo "Detected a license file in /var/lib/rstudio-connect/*.lic."
```
dbkegley, line 40: This prevents the following `elif` block from executing which would mean that the license file specified by `$RSC_LICENSE_FILE_PATH` won't be activated

ianpittwood: That's a fair point, but wouldn't any file already in `/var/lib/rstudio-connect/*.lic` activate automatically too on startup? Would the preferred method be to do an `rm /var/lib/rstudio-connect/*.lic` if `RSC_LICENSE_FILE_PATH` is defined?

bschwedler (approved):

connect/startup.sh
```diff
@@ -37,14 +37,16 @@ elif ! [ -z "$RSC_LICENSE_SERVER" ]; then
     /opt/rstudio-connect/bin/license-manager license-server $RSC_LICENSE_SERVER
     trap deactivate EXIT
 elif test -f "$RSC_LICENSE_FILE_PATH"; then
-    rm -f /var/lib/rstudio-connect/*.lic
-    cp "${RSC_LICENSE_FILE_PATH}" /var/lib/rstudio-connect/license.lic
-    chmod g-rwx,g-rwx /var/lib/rstudio-connect/license.lic
+    rm -f /var/lib/rstudio-connect/*.lic || true
+    /opt/rstudio-connect/bin/license-manager activate-file $RSC_LICENSE_FILE_PATH
```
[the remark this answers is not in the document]
aronatkins: The installed ownership and permissions may not be accurate when using `activate-file`.

Follow the guidance provided by each product. Depending on how the product is installed and if run by a privileged user, the requirements are different.

Please do not revert the changes I made in https://github.com/rstudio/rstudio-docker-products/pull/927.

PR #938, opened 2025-06-23
content: An image with R 4.5, Python 3.13, Quarto 1.7, and latest drivers

aronatkins:
Merges to main. A new image having R 4.5.

@bschwedler & @ianpittwood - I've updated the drivers version in this new image but not elsewhere. It looks as if we have a variety of drivers version references. Should I use the 2024.03.0 drivers like all of the other images?

The `Justfile`, for example, references 2023.05.0, while there are many other references to 2024.03.0. My change is the only reference to 2025.03.0, as far as I can tell.


bschwedler (commented):
We should always try to use the latest released version of the drivers.

The bakefile(s) controls the versions that are installed in the published images.
Historically, the default `ARG` values in each `Dockerfile` have not been kept well up-to-date.

If you don't anticipate it causing issues with the content images, we can also update the driver version in both bakefiles. This could be done in a separate PR.

Perhaps it's time to remove these `Dockerfile` defaults in a separate PR to prevent similar confusion in the future.

docker-bake.hcl
```diff
@@ -161,6 +161,9 @@ variable CONTENT_BUILD_MATRIX {
 
       # R-4.4, Python-3.12.4, Quarto-1.4.557 (polyfill.js vulnerability patch)
       {os = "ubuntu2204", os_alt = "jammy", r = "4.4.1", py = "3.12.4", drivers = "2024.03.0", quarto = "1.4.557"},
+
+      # R-4.5, Python-3.13, Quarto-1.7
+      {os = "ubuntu2204", os_alt = "jammy", r = "4.5.1", py = "3.13.5", drivers = "2025.03.0", quarto = "1.7.32"},
```
bschwedler, line 166: Let's keep this consistent with the existing content images for now. We can upgrade them all in a separate PR.
```suggestion
      {os = "ubuntu2204", os_alt = "jammy", r = "4.5.1", py = "3.13.5", drivers = "2024.03.0", quarto = "1.7.32"},
```

bschwedler (changes requested):

docker-bake.hcl
aronatkins: Done.

aronatkins:
> LGTM. Out of curiosity, should we start producing these on an Ubuntu 24 base now too?

I am in favor of using Ubuntu 24 as a base, but that should happen alongside these image. In my opinion, your in-flight work to restructure image building (and names) is the right place to adjust the base image.

PR #943, opened 2025-07-03
chore: increase RSW_LAUNCHER_TIMEOUT from 10 to 60

fh-mthomson:
Per Posit Support Ticket #107355, we've seen our Kubernetes Workbench pod get stuck in a crashloop (more than >20% of the time), due to "wait-for-it.sh" timing out after 10 seconds: https://github.com/rstudio/rstudio-docker-products/blob/3bb8d5f9835f30ed93b00f92005b114ebeb598f1/workbench/startup.sh#L86

Example error (redacted slightly, for illustration):
```
wait-for-it.sh: timeout occurred after waiting 10 seconds for localhost:5559
== Exiting ==
....
Deactivating license ...
...
```

We've confirmed that bumping `RSW_LAUNCHER_TIMEOUT` to 60s is plenty to ensure the launcher starts up in time.

Feel free to modify as applicable - we're not ultimately familiar with the intracies or intent of `wait-for-it.sh` (since it's not part of the standard Workbench setup, afaik).


fh-mthomson:
@ianpittwood @bschwedler are y'all open to reviewing/including this change before next Workbench release?

bschwedler:
@fh-mthomson Have you tried adding the `RSW_LAUNCHER_TIMEOUT` env var using Helm values?

```yaml
pod:
  # -- env is an array of maps that is injected as-is into the "env:" component of the pod.container spec
  env:
    - name: RSW_LAUNCHER_TIMEOUT
      value: 60
```
https://github.com/rstudio/helm/blob/c7edade7458d2122142525d56c03098c8c308c10/charts/rstudio-workbench/values.yaml#L366-L367

This will interact with the [startupProbe](https://github.com/rstudio/helm/blob/c7edade7458d2122142525d56c03098c8c308c10/charts/rstudio-workbench/values.yaml#L278-L288), _but_ the default total timeout for the probe is `300s`.

---
We can also update the default in the container images.

I'll have to cherry-pick your change into a separate PR.  Builds from forks will still fail until we are able to overhaul the image workflows.

fh-mthomson:
> Have you tried adding the `RSW_LAUNCHER_TIMEOUT` env var using Helm values?

We considered it and agree it would work! 

But felt it was slightly more self-contained to override the env var within the specific Docker image from which it was being set/used (via `wait-for-it.sh` in the upstream Posit image, rather than the Workbench app itself).

bschwedler:
Sounds good.

I've created #958 so that this can run through CI.  I've set it to 30s. We've been using this value in our internal Kubernetes clusters for some time.

fh-mthomson:
Amazing, thank you! Always helpful to know what has worked well - 60s was a conservative guess.

PR #948, opened 2025-07-15
Add load flag to docker/bake-action

bschwedler:
This is an attempt to fix inconsistent output of layer digests.

`docker history` was not showing the digets output for images pulled
from Docker Hub or GHCR.

This issue indicates that in order for buildx to output the digests, we
need to set `load: true`
https://github.com/docker/build-push-action/issues/906


ianpittwood (approved):
Should we bump to the v6 action too since it's out?

bschwedler:
> Should we bump to the v6 action too since it's out?

Good call. I'll pull that in and rebase.

PR #951, opened 2025-08-05
Use default Python install, not jupyterlab install

cm421:
Brought up in a customer ticket #118720 pointing out that the `jupyter` version of python isn't user-writable so the default python version for this image isn't usable out of the box. 

```
$ which python
/opt/python/jupyter/bin/python
$ pip install scipy
...
Installing collected packages: numpy, scipy
ERROR: Could not install packages due to an OSError: [Errno 13] Permission denied: '/opt/python/jupyter/lib/python3.12/site-packages/numpy'
Check the permissions.
```

This reverts a change to set whatever the `PYTHON_VERSION` is to be the default python version

### DIFF
--- a/r-session-complete/Dockerfile.ubuntu2204
+++ b/r-session-complete/Dockerfile.ubuntu2204
@@ -58,7 +58,7 @@ RUN /opt/python/"${PYTHON_VERSION}"/bin/python -m venv /opt/python/jupyter \
     && /opt/python/${PYTHON_VERSION_ALT}/bin/python -m ipykernel install --name py${PYTHON_VERSION_ALT} --display-name "Python ${PYTHON_VERSION_ALT}" \
     && /opt/python/jupyter/bin/python -m pip cache purge
 
-ENV PATH="/opt/python/jupyter/bin:${PATH}"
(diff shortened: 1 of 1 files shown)


cm421:
I'm not sure why the tests are failing on quarto as my change shouldn't have bumped that. Though I see a lot of other PRs with failing tests so I'm going to assume that the failure isn't specific to my changes

cm421:
Thank you both! I've updated to keep both Jupyter and the default python version on PATH. Curious to see if maybe Quarto did want some python packages that were in the Jupyter install?

cm421:
@bschwedler sorry for the delay, is using `QUARTO_PYTHON` like this what you were thinking? I'm not sure if it's supposed to be used with a venv like this but it dos build and it passes `quarto check` without issue and also allows users to install packages with the normal python install.

bschwedler:
> What surprises me with the user I created is that the `PATH` with the full python install was clobbered by the user's default shell/profile, so the python wasn't readily available. I thought this _was_ working before, where setting `PATH` with quarto's python in the front was getting picked up in user sessions.
> 
> Does this seem like what you'd expect?

Perhaps we need to add this to `/etc/paths` as well?

PR #959, opened 2025-08-14
Prep for `dev` branch removal

bschwedler:
- **Remove @costrouc from CODEOWNERS**
- **Remove `SRC_IMAGE_NAME` from ci.justfile**
- **Remove unused just recipes from ci.justfile**
- **Remove separate `dev` branch treatment in GitHub Actions**
- **Update NEWS.md to remove the `dev` branch**


ci.Justfile
ianpittwood, line 1: Should we consider updating the `Justfile` too with any removals and the merge them together?

.github/workflows/build-bake.yaml
ianpittwood, line 1: We should be able to remove the split checkouts too, but only after the primary branch is swapped for `main`.

There are also several instances of `${{ github.ref == 'refs/heads/main' || github.event_name == 'schedule' }}` for push that we can remove `github.event_name == 'schedule'` from when `main` is the primary.

ianpittwood (approved):
I think this is OK as is, just a few comments of things you're probably already aware of. Thanks for doing this :D

ci.Justfile
bschwedler: That's a good idea. Let me do that and push another commit.

PR #981, opened 2025-10-15
Add soci v2 index to workbench-session-init

jforest:
Not the most elegant thing.  `ctr` doesn't pick up regular docker credentials, that's why the hacky long `if`/`elif`/`else` line


ianpittwood (approved):
LGTM! Thanks for doing this!

PR #984, opened 2025-10-31
Bump workbench version to 2025.09.2

jeffvroom:


adamhigerd (approved):
Surely there must be a way to have some of these pull the information in from a single source instead of having so many redundant tweaks. Oh well. An optimization for another time.

PR #985, opened 2025-11-05
Remove Dockerfile `ARG` instruction defaults

ianpittwood:
- Remove default `ARG` values from all Dockerfiles in favor of only using values specified in `docker buildx bake` definitions.
  - Remove old `sed` commands related to updating these values from the `Justfile`.
  - Slight tweaks and cleanup for version definitions in `docker-bake.hcl`.
- Remove unused files such as `matrix-preview.json` and `.env` files.
- Update Connect Python, R, and Quarto versions since they lag significantly behind and 2025.10.0 was only just released last week.
- Update WAML Python and R versions.
  - Remove WAML Azure ML SDK v1 package from Python installs, no longer compatible.
- Update all preview/daily build Python and R versions.
- Update `NEWS.md` and `README.md` files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,9 +126,10 @@ We orchestrate all our builds using `docker buildx bake`. You can learn more abo
 [Docker's buildx bake page](https://docs.docker.com/build/bake/), however no additional background knowledge is needed on the tool in order 
 to use it.
 
-All dependency versions (Python, R, Quarto, etc.) are defined in the `docker-bake.hcl` file at the root of the
-repository. You can change versions there to suit your needs. To get a JSON representation of all targeted image builds 
-and their build arguments, call the `just plan` target.
+> [!NOTE]
+> All dependency versions (Python, R, Quarto, etc.) are defined in the `docker-bake.hcl` file at the root of the
+> repository. You can change versions there to suit your needs. To get a JSON representation of all targeted image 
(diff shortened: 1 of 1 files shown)


Justfile
```diff
@@ -271,11 +248,6 @@ update-rsw-versions:
 update-rspm-versions:
   #!/usr/bin/env bash
   set -euxo pipefail
-  sed {{ sed_vars }} "s/RSPM_VERSION=.*/RSPM_VERSION={{ RSPM_VERSION }}/g" \
-    package-manager/.env \
-    package-manager/Dockerfile.ubuntu2204
-  sed {{ sed_vars }} "s/RSPM_VERSION:.*/RSPM_VERSION: {{ RSPM_VERSION }}/g" docker-compose.yml
-  sed {{ sed_vars }} "s/rstudio\/rstudio-package-manager:.*/rstudio\/rstudio-package-manager:$(just _get-clean-version {{ RSPM_VERSION }})/g" docker-compose.yml
   sed {{ sed_vars }} "s/^RSPM_VERSION := .*/RSPM_VERSION := \"{{ RSPM_VERSION }}\"/g" \
     Justfile
```
bschwedler: Do we need to update the `Justfile` if we removed it as a variable from the beginning of the file?

bschwedler (approved):

PR #993, opened 2026-01-07
Bump Workbench to 2026.01.0+392.pro5

skyeturriff:


bschwedler:
@skyeturriff I don't know exactly why it's failing, but it looks like the container builds successfully but fails at the `test` step:

https://github.com/rstudio/rstudio-docker-products/blob/95d293d184767a8b98892e63ab6e1367c3aaadff/.github/actions/bake-test-push/action.yml#L151-L156

The tests should use the image that was built locally on the same GHA runner and tests need to pass before it can be pushed to the container registries.


I thought it might be silently failing in an earlier stage so I created #994 stacked on top of this and the first build passed.

I re-ran the workflow here and it succeeded. I'm wondering if there was a timing issue where the session components weren't available?

bschwedler (approved):

bschwedler:
Merging. We can investigate AzureML separately.

PR #1008, opened 2026-02-25
Temporarily disable TensorFlow Serving installation

ianpittwood:


millerick:
We manually disabled tensorflow in our configuration since we weren't using it, but it isn't great that your build process pushed updated versions of existing docker image tags with this removed.  Users shouldn't need to upgrade to a new version of your helm chart to get around an issue you introduce to gloss over failures in your image build process.

aronatkins:
@millerick - We understand the frustration caused by this change. We're sorry that this affected you.

Unfortunately, because of a problem (https://github.com/tensorflow/serving/issues/4131) with the TensorFlow serving project, we are no longer able to download and install that binary into the Connect image. We needed to remove TensorFlow before that image could build again.

TensorFlow was removed ahead of the Connect 2026.02 release, disturbing existing releases. That was not our intention and a mistake on our part. The TensorFlow change should have accompanied the Connect 2026.02.0 release without modifying existing published images.

The in-image product configuration and Helm chart values both referenced the `TensorFlow.Executable` binary that no longer existed.

You're right that you shouldn't have needed to update the Helm chart. That said, we also neglected to update the Helm chart as this release went live. That has now been resolved.

As @tdstein mentioned, we're including these problems in the retrospective discussion we run after each product release.

PR #1041, opened 2026-05-11
Update TinyTeX install URL to tinytex.yihui.org

dotNomad:
## Summary

Updates `content/base/Dockerfile.ubuntu2204` to fetch the TinyTeX install script from `tinytex.yihui.org` directly instead of going through `yihui.org/tinytex/install-bin-unix.sh`.

## Why

Per the [TinyTeX FAQ](https://yihui.org/tinytex/faq/), the canonical install URL has moved to `tinytex.yihui.org`. The old URL is now a 301 redirect to the new one, but the redirect host has been returning intermittent HTTP 500s — which fails the content-base build under `curl -fsSL`'s strict error handling.

Verified locally:

```
$ curl -sI https://yihui.org/tinytex/install-bin-unix.sh | head -1
HTTP/2 301
$ curl -sI https://tinytex.yihui.org/install-bin-unix.sh | head -1
HTTP/2 200
```

Pointing directly at the new host removes the redirect hop and the flaky old host from the critical path.

### DIFF
--- a/content/base/Dockerfile.ubuntu2204
+++ b/content/base/Dockerfile.ubuntu2204
@@ -88,7 +88,7 @@ RUN apt-get update \
 RUN apt-get update  \
     && apt-get install -y --no-install-recommends wget \
     && rm -rf /var/lib/apt/lists/* \
-    && curl -fsSL "https://yihui.org/tinytex/install-bin-unix.sh" | sh \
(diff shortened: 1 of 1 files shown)


dotNomad:
Closing since https://github.com/posit-dev/images-connect already has a way to address this.

PR #1052, opened 2026-06-05
Schedule end-of-2026 deprecation in all READMEs

bschwedler:
All images deprecated at end of 2026 now print a warning with a direct
migration link when the container starts. Previously the deprecation
notice only appeared in the README.

For images that already had shell startup scripts (connect, workbench,
package-manager, connect-content-init), the warning is prepended to the
existing script. For images that ran a binary as CMD
(workbench-session-init, workbench-positron-init) or had no entrypoint
at all (workbench-session, r-session-complete, content/base, content/pro),
a new entrypoint.sh wrapper is added that prints the warning and exec's
the original command.

Images that inherit from product-base-pro (r-session-complete,
workbench-session) chain through tini to preserve PID 1 signal handling:
`ENTRYPOINT ["/tini", "--", "/entrypoint.sh"]`.

Migration links point to the individual image README in the replacement
repo (posit-dev/images-connect, images-workbench, images-package-manager).


kfeinauer (approved):

bschwedler:
@ianpittwood Do you think we should log this in the startup/entrypoint as well?

ianpittwood:
> @ianpittwood Do you think we should log this in the startup/entrypoint as well?

A warning would be great! I'm sure many use the containers without checking the README.

ianpittwood (dismissed):
LGTM, timeline is good. I like the suggestion of putting an extra warning in the startup script though.

bschwedler:
@ianpittwood Could I please have you check the entrypoint modifications? Thanks!

ianpittwood (approved):
Looks good content-wise, but it would be preferable to write to `stderr` instead of `stdout`.

connect/README.md
```diff
@@ -1,9 +1,9 @@
 # Quick reference
 
-> [!NOTE]
+> [!IMPORTANT]
 > A new <a href="https://github.com/posit-dev/images-connect/blob/main/connect/README.md">posit/connect</a> image is now generally available and replaces this image.
 >
-> - This image continues to be maintained. A deprecation date will be announced separately.
+> - Posit will deprecate this image at the end of 2026. Version updates will continue until then. Please migrate to the Posit image linked above.
```
edavidaja: "Will deprecate"? I think this is the deprecation notice and we weill stop publishing updates at the end of the year?

bschwedler: Good call. I've updated the wording to:
> Posit will stop publishing this image at the end of 2026.

